PR #3, opened 2019-02-07
Using a configurable properties file

seanmcilvenna:
Using a configurable properties file for settings that might change between various environments.


src/main/java/ca/uhn/fhir/jpa/demo/FhirServerConfig.java
```diff
@@ -62,41 +64,47 @@ public DatabaseBackedPagingProvider databaseBackedPagingProvider() {
...
 	@Override
 	@Bean()
 	public LocalContainerEntityManagerFactoryBean entityManagerFactory() {
 		LocalContainerEntityManagerFactoryBean retVal = super.entityManagerFactory();
-		retVal.setPersistenceUnitName("HAPI_PU");
-		retVal.setDataSource(dataSource());
+		retVal.setPersistenceUnitName(HapiProperties.getPersistenceUnitName());
+
+		try {
+			retVal.setDataSource(dataSource());
```
jamesagnew: It'd probably make more sense to just re-throw the exceptions below, given that we won't be able to start without this succeeding

src/main/java/ca/uhn/fhir/jpa/demo/HapiProperties.java
```diff
@@ -0,0 +1,211 @@
...
+            // Load the configurable properties file
+            InputStream in = null;
+
+            try {
+                in = HapiProperties.class.getClassLoader().getResourceAsStream(HAPI_PROPERTIES);
+                HapiProperties.properties = new Properties();
+                HapiProperties.properties.load(in);
+                in.close();
+            } catch (FileNotFoundException e) {
+                e.printStackTrace();
```
jamesagnew: Same deal here, I'd just re-throw these.

seanmcilvenna: I've updated to have handled exceptions re-throw the exceptions as ConfigurationExceptions.

seanmcilvenna:
I don't do a lot of pull requests... I think I addressed your concerns.
Is there something else I need to do??

PR #4, opened 2019-02-11
Improve multiversion support

jamesagnew:
@seanmcilvenna would you ming giving this a revue? I've built on your awesome work with the properties file to get it to the point where switching FHIR versions is just a setting in the file and no other changes required.

### DIFF
--- a/src/main/resources/hapi.properties
+++ b/src/main/resources/hapi.properties
@@ -1,7 +1,19 @@
+
+# Adjust this to set the version of FHIR supported by this server. See
+# FhirVersionEnum for a list of available constants.
 fhir_version=DSTU3
+
+# This is the address that the FHIR server will report as its own address.
+# If this server will be deployed (for example) to an internet accessible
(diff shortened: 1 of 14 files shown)


src/main/resources/hapi.properties
```diff
@@ -1,7 +1,19 @@
+
+# Adjust this to set the version of FHIR supported by this server. See
+# FhirVersionEnum for a list of available constants.
 fhir_version=DSTU3
```
seanmcilvenna: is src/main/resources/hapi.properties the right place to put the properties file so that the user can deploy the WAR file in tomcat, and then update the properties file, and have subsequent deploys not destroy/overwrite the properties file???

jamesagnew: @seanmcilvenna good question.

Another thing on my mind- I'd like to configure this project to build both a WAR file as well as an executable JAR (with embedded HTTP server) so that it can be run without needing to install in a container.

PR #6, opened 2019-02-11
More properties

seanmcilvenna:


jamesagnew:
This is such an awesome strategy for configuring this project. I'm really loving the direction you're taking it.

PR #12, opened 2019-02-20
Docs of hapi-fhir-testpage-overlay not loading properly due to local conformance statement issue

jjcosare:
- fixed issue on local conformance statement not loading properly


seanmcilvenna:
Why do you say it's not loading propertly? I run it as an "Application" within IntelliJ with the default settings without a prob... Are you running this via Jetty? or via Tomcat?

jjcosare:
> Why do you say it's not loading propertly? I run it as an "Application" using the ExampleServerDstu3IT class as the startup/main class within IntelliJ with the default settings without a prob... Are you running this via Jetty? or via Tomcat?

Running it via jetty based on docs `mvn jetty:run` and it doesn't load the hapi-fhir-testpage-overlay based on default settings... Upon checking more closely on ExampleServerDstu3IT, it doesn't have the same contextPath set based on the project's jetty plugin pom.xml so we end up having different results... Updating the contextPath of ExampleServerDstu3IT to "/hapi-fhir-jpaserver" will reveal the issue for us

src/main/java/ca/uhn/fhir/jpa/starter/JpaRestfulServer.java
```diff
@@ -170,21 +172,42 @@ protected void initialize() throws ServletException {
             registerProvider(retriggeringProvider);
         }
 
+        // Define your CORS configuration. This is an example
+        // showing a typical setup. You should customize this
+        // to your specific needs
+        CorsConfiguration config = new CorsConfiguration();
```
jamesagnew: It'd probably make sense for this to be exposed in the HapiProperties file as a config item (maybe with the allowed origins being modifiable?)

src/main/resources/hapi.properties
```diff
@@ -31,14 +31,14 @@ logger.name=fhirtest.access
 logger.format=Path[${servletPath}] Source[${requestHeader.x-forwarded-for}] Operation[${operationType} ${operationName} ${idOrResourceName}] UA[${requestHeader.user-agent}] Params[${requestParameters}] ResponseEncoding[${responseEncodingNoDefault}]
 logger.error_format=ERROR - ${requestVerb} ${requestUrl}
 logger.log_exceptions=true
-datasource.driver=org.apache.derby.jdbc.EmbeddedDriver
-datasource.url=jdbc:derby:directory:target/jpaserver_derby_files;create=true
-datasource.username=
-datasource.password=
+datasource.driver=com.mysql.cj.jdbc.Driver
```
jamesagnew: I would rather leave the default configuration using Derby since that will work for anyone regardless of what other software they have installed.

src/main/java/ca/uhn/fhir/jpa/starter/JpaRestfulServer.java
patrick-werner: This would be great.

src/main/resources/hapi.properties
jjcosare: Noted, have reverted back to derby and instead added some updates on readme regarding mysql usage.

src/main/java/ca/uhn/fhir/jpa/starter/JpaRestfulServer.java
jjcosare: Thank you, have updated it to be configurable based on your suggestion. Also removed the unused lib on pom.xml and cors configuration on web.xml.

jamesagnew:
Looks great! Merging now.

PR #14, opened 2019-03-04
Add cache timeout configuration property. Disabled by default

johnpoth:
Adds cache timeout configuration property.

This is useful when using hapi-fhir-jpaserver-starter as a test FHIR server which relies on returned results not being stale.

I disabled caching by default.

Thanks!


src/main/java/ca/uhn/fhir/jpa/starter/HapiProperties.java
```diff
@@ -14,6 +14,7 @@
     static final String ALLOW_EXTERNAL_REFERENCES = "allow_external_references";
     static final String ALLOW_MULTIPLE_DELETE = "allow_multiple_delete";
     static final String ALLOW_PLACEHOLDER_REFERENCES = "allow_placeholder_references";
+    static final String CACHE_TIMEOUT = "cache_timeout";
```
jamesagnew: Could this be called `search_result_cache_timeout_millis` just to make it a bit more explicit what it is?

johnpoth: totally agree! I changed it to the same name as the config dao so `reuse_cached_search_results_millis` if that's cool and rebased so no more conflicts from master. Thanks!

PR #18, opened 2019-03-19
Schema-base multi-tenancy example

col-panic:
Concerning the Support for Multitenancy  [google-group discussion](https://groups.google.com/forum/#!topic/hapi-fhir/F37kfAHYXnc) I implemented a concise example - this PR.

I would very much appreciate any comments on the code done here. It is not a prior target to get this PR included - but to have an appropriate discussion point for persons interested in this functionality.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,3 +63,25 @@ To configure the starter app to use MySQL, instead of the default Derby, update 
 * hibernate.dialect=org.hibernate.dialect.MySQL5Dialect
 
 It is important to use MySQL5Dialect when using MySQL version 5+.
+
+## Multi-Tenancy
+
+To enable schema-based multi-tenancy, change the following settings:
+
+* Switch to MySQL service, as described above, and create a database `hapi_test` (the default database)
+* For each tenant to use, create an additional database `hapi_tenantid`
+* Import the following [schema](https://bitbucket.org/hspconsortium/reference-api/src/d5d7eb46d2399213633e9f6de0415774871321f3/reference-api-mysql/src/main/resources/db/mysql/hspc_8_schema_empty.sql?at=develop "hspc_8_schema_empty.sql") to each of the `hapi_*` databases
+* In `hapi.properties` enable and configure all elements in the `Multi-Tenancy` section
+* In order to perform a query on specific tenant, add the HTTP-Header `X-TenantID` to your request.
+
+Relevant discussions on this topic:
+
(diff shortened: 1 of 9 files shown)


jamesagnew:
This is really neat! Nice work.

I guess my main comment would be that it would be preferable to replace the tenant identification Filter with an implementation of `ITenantIdentificationStrategy`. This has a couple of advantages:

* It is the HAPI FHIR native way of doing multitenancy, which means it'll play nicely with `AuthorizationInterceptor` and potentially other interceptors that are aware of it
* It can be registered with the server at runtime, meaning that it could be enabled purely based on properties in the `hapi.properties` file

One other thought- it'd probably be good for this to have a unit test since it seems like something that will fail into disrepair pretty fast if it's not tested regularly. That's tough though of course, since that one has to work on Derby and I don't think Derby supports the `USE` clause you're using with MySQL.

Presumably we'd have to use a separate schema (i.e. a different JDBC URL) per tenant on Derby..

col-panic:
Thanks for your comment!

I don't see how FHIR itself, resp. `ITenantIdentificationStrategy` would fit to this scenario, if the "tenant-switch" is already done on JDBC resp. connection level. HAPI-FHIR shouldn't even bother here, no? I mean every request just operates against a totally different scheme.

I see that this patch is only working for mysql by now, for derby you would have to have a separate connection per tenant, which you then hand out. I'll take a look whether I can augment the starter with this scenario.

If HAPI-FHIR was to do the tenant handling, wouldn't it be only able to do single-database-multitenancy by means of tenant discriminators such that the SQL requests may consider them? `ITenantIdentificationStrategy` currently does not offer me any possibility to replace the underlying JDBC connection - but only to change the servers access url.

Unit tests - certainly - as soon as all discussion is done ;)

jkiddo:
@col-panic / @jamesagnew is this still relevant?

col-panic:
@jkiddo this was a prototype for a project that was not done. So for me it is not relevant, maybe it would be interesting to keep it if another developer runs into this?

tsun424:
Hi @col-panic , thanks for creating this PR and keeping it open for other developers. I wonder if you've ever implemented this schema-based multi-tenancy solution in any production environment, if you did, have you had any issues for this solution? Thank you.

patrick-werner:
as this was only a prototype, im closing this PR

PR #19, opened 2019-03-21
Sample data generator with Synthea

jjcosare:
Added synthea script for automation on adding data into fhir server, this would be useful in testing out the starter project with some dummy data

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,3 +74,25 @@ FLUSH PRIVILEGES;
 * hibernate.dialect=org.hibernate.dialect.MySQL5Dialect
 
 It is important to use MySQL5Dialect when using MySQL version 5+.
+
+## Synthea<sup>TM</sup> Patient Generator 
+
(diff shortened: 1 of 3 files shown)


patrick-werner:
i like the idea in general, but i personally think this should be part of the synthea distribution and not part of the jpaserver-starter.

seanmcilvenna:
@patrick-werner I agree. However, I don't want to outright reject the PR. Will leave that to @jamesagnew

seanmcilvenna:
@jjcosare there are conflicting files... I'm still leaving @jamesagnew to review this, but thought I would point out that it definitely won't be merged until at least the conflicts are resolved :)

psavva (commented):
I would recommend a widows batch file to be included with would be on par with the file bash file clean_generate_install_fhir.sh

jkiddo:
Closed due to inactivity

PR #20, opened 2019-03-31
Add support for websockets

jamesagnew:


jamesagnew:
@seanmcilvenna Would you mind to give this a review?

This PR has the following changes:

* Make the persistence unit name non-configurable (I just can't think of a reason anyone would want to change this so it seemed misleading to make it configurable)
* Add support for Websocket subscriptions
* Change the context path used by the mvn Jetty plugin to use `/hapi-fhir-jpaserver` just so that it will be consistent with what people get when they deploy to Tomcat. Hopefully this will prevent some confusion by newbies.

PR #23, opened 2019-05-01
Pom mail

seanmcilvenna:
Adding javax.mail dependency to pom.xml so that when properties are configured to enable mail, the pom.xml doesn't have to be additionally modified for it to work.


jamesagnew:
Looks good!

PR #29, opened 2019-05-28
changed default to R4 fixes #28

### LINKED ISSUE #28: properties file should configure R4 as default
currently this project has R3 as default, as R4 is normative we should change the default to a valid R4 config.

patrick-werner:


pom.xml
```diff
@@ -53,7 +53,7 @@
         <!-- At least one "structures" JAR must also be included -->
         <dependency>
             <groupId>ca.uhn.hapi.fhir</groupId>
-            <artifactId>hapi-fhir-structures-dstu3</artifactId>
```
seanmcilvenna: @jamesagnew can you confirm if this change is needed? I've built jpaserver-starter for R4 without changing this in the pom.xml before...

patrick-werner: you need the corresponding structures package to the fhir_version in the properties file, otherwise you will get runtime errors.

seanmcilvenna: Oddly, I have not experienced that... I have an R4 jpaserver-starter server running right now that hasn't complained about the structures be the wrong version. I hadn't noticed this line in pom.xml, so I didn't change it when I built my R4 server...

patrick-werner: you are right, all fhir versions are already included in jpaserver-base, import was unnecessary --> removed

PR #38, opened 2019-07-04
Updated readme file with option to change port

quintonn:
Updated readme file with instructions on how to run the hapi server on a different port using Maven


jamesagnew:
Fantastic! Thanks!

PR #36, opened 2019-06-20
Containerize maven build process

ivan-c:
I don't typically keep the Java toolchain installed locally and found these changes convenient.

Please let me know if you have any thoughts!

---
* Run maven build process inside prerequisite [docker stage](https://docs.docker.com/develop/develop-images/multistage-build/)
* Remove `build-docker-image.sh` wrapper (only `docker build` is required now)
* Update base images
  * Update to JDK 11 in jetty base image
  * Use Debian-based base images (instead of Alpine) for least developer surprise

PR #42, opened 2019-08-02
Building app in docker. Easier to override config file from /hapi-con…

seanmcilvenna:
…fig/hapi.properties


seanmcilvenna:
I think this is actually very similar to https://github.com/hapifhir/hapi-fhir-jpaserver-starter/pull/36
I didn't realize https://github.com/hapifhir/hapi-fhir-jpaserver-starter/pull/36 existed. But, I think it may be a little better because of the abstraction of the config file, which can be overridden when running the docker image.

PR #45, opened 2019-08-07
Cors fix

patrick-werner:
Using cors.allowed.origin=* together with credentials is forbidden:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Server-Side_Access_Control
https://www.w3.org/TR/cors/

and results in a 403 response.
This PR adds support for allowCredentials, and a list of allowed origins.
The spring CORS filter will check if the origin is in this list an returns the appropriate url.

### DIFF
--- a/src/main/java/ca/uhn/fhir/jpa/starter/HapiProperties.java
+++ b/src/main/java/ca/uhn/fhir/jpa/starter/HapiProperties.java
@@ -42,7 +42,7 @@ public class HapiProperties {
     static final String TESTER_CONFIG_REFUSE_TO_FETCH_THIRD_PARTY_URLS = "tester.config.refuse_to_fetch_third_party_urls";
     static final String CORS_ENABLED = "cors.enabled";
     static final String CORS_ALLOWED_ORIGIN = "cors.allowed_origin";
-    static final String CORS_ALLOWED_CREDENTIALS = "hapi.properties";
+    static final String CORS_ALLOW_CREDENTIALS = "cors.allowCredentials";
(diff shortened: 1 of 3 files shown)


src/main/java/ca/uhn/fhir/jpa/starter/HapiProperties.java
```diff
@@ -42,6 +42,7 @@
     static final String TESTER_CONFIG_REFUSE_TO_FETCH_THIRD_PARTY_URLS = "tester.config.refuse_to_fetch_third_party_urls";
     static final String CORS_ENABLED = "cors.enabled";
     static final String CORS_ALLOWED_ORIGIN = "cors.allowed_origin";
+    static final String CORS_ALLOWED_CREDENTIALS = "hapi.properties";
```
jamesagnew: I guess this should probably be `CORS_ALLOW_CREDENTIALS` (no 'ed') and I think the value is wrong :)

patrick-werner: you are right. Stupid minor mistakes. Fixed it

patrick-werner:
@jamesagnew would love to merge, but i still don't have contributor rights on this repo.

jamesagnew:
@patrick-werner - Can you check the email account you have registered with GitHub? You should see an invite.

patrick-werner:
i got the invite, but i don't have rights within this project.

jamesagnew:
Can you try again? I'm still learning how permissions work in GitHub after all these years apparently :p

patrick-werner:
it worked. Thanks @jamesagnew

PR #72, opened 2019-12-20
Start 4.2.0 release branch

jamesagnew:


seanmcilvenna:
question about process...
should new version (XXX-SNAPSHOT) stuff be done in a separate branch until it is completed, and then master represents a version of "4.2.0" (NOT snapshot)?
just thinking that the master branch should always have the latest stable release...

jamesagnew:
@seanmcilvenna Yup, this won't be merged until 4.2.0 is released (guess I should probably have commented that here). I like to have a branch going though so that people can try out snapshot if they want, and so that I can run CI on this project against snapshot hapi

seanmcilvenna:
Makes sense. Sounds great!

PR #80, opened 2020-02-03
3247/durant/postgres container no fhirbase

durantschoon:
First tests to see if the server and postgres instance are running.

Should trigger travis so check to verify this command is indeed run by travis:

`./src/test/bash/hapifhir_postgres/test_all.sh run_server`


durantschoon:
upstream mistake, sorry. Trying to delete

PR #84, opened 2020-03-01
Update hapi.properties

neba123:


patrick-werner:
This PR propably was sent to the wrong remote.

PR #99, opened 2020-05-14
Adding environment variables to override hapi properties file settings

petermicuch:
I am happy with most of the hapi properties :-). There is today possibility to override hapi properties by   `-Dhapi.properties` in `JAVA_OPTS` environment variable.
However it is very convenient to just override some of the properties in your container using kubernetes manifest file for your deployment. Like in this example, where I override default DB used by HAPI and set desired version.
```
        env: 
        - name: HAPI_DATABASE_NAME
          value: HAPI_R4
        - name: datasource.url
          value: "jdbc:sqlserver://mssql-database;databaseName=$(HAPI_DATABASE_NAME)"
        - name: datasource.driver
          value: com.microsoft.sqlserver.jdbc.SQLServerDriver
        - name: hibernate.dialect
          value: org.hibernate.dialect.SQLServer2012Dialect
        - name: server_address
          value: {{http-scheme}}://{{base-url}}{{hapi-r4-base-uri}}
        - name: fhir_version
          value: R4
        - name: datasource.username
          valueFrom:
            secretKeyRef:
              name: hapi-db-secret-name
              key: username  
        - name: datasource.password
          valueFrom:
            secretKeyRef:
              name: hapi-db-secret-name
              key: password 
```

So instead of keeping and mounting different hapi.properties files or having customized files build into docker image, I can use one default docker image and customize in kubernetes manifest files as needed. I.e. running DSTU3 and DSTU2 version side by side using the same container. No matter what solution you use to maintain your manifest files (helm, kustomize...), this works.

To have that possibility, environment variables have to be loaded and override properties from hapi.properties file. Solution provided here is very simple and works, but I guess could be improved e.g. by adding only variables that are "known" in hapi.properties file. I am actually using image with this change and it would be great, if this could be part of default source code.

### DIFF
--- a/src/main/java/ca/uhn/fhir/jpa/starter/HapiProperties.java
+++ b/src/main/java/ca/uhn/fhir/jpa/starter/HapiProperties.java
@@ -17,6 +17,7 @@ import java.util.Arrays;
 import java.util.Locale;
 import java.util.Properties;
 import java.util.Set;
+import java.util.Map;
 import java.util.stream.Collectors;
 
 import static org.apache.commons.lang3.StringUtils.defaultString;
@@ -138,7 +139,10 @@ public class HapiProperties {
     if (overrideProps != null) {
       properties.putAll(overrideProps);
     }
-    properties.putAll(System.getenv());
+    properties.putAll(System.getenv().entrySet()
+                                     .stream()
+                                     .filter(e -> e.getValue() != null && properties.containsKey(e.getKey()))
+                                     .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue)));
     return properties;
   }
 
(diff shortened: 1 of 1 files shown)


AdrianLsk:
Hi, I have an off-topic question.

Did you get the MS SQL working without a problem? 
When I try your hapi.properties I am getting the following error:

```Caused by: 
javax.persistence.PersistenceException: [PersistenceUnit: HAPI_PU] Unable to build Hibernate SessionFactory; nested exception is java.lang.IllegalStateException: Connection factory returned null from createConnection
```

petermicuch:
Yes, getting MSSQL running is not an issue. You can not just use my hapi.properties. You need to change pom file to include mssql connector and build with that.
```
<dependency>
	<groupId>com.microsoft.sqlserver</groupId>
	<artifactId>mssql-jdbc</artifactId>
	<version>7.4.1.jre8</version>
</dependency>
```
And of course you need to have correct connection string.

AdrianLsk:
Thanks for your reply.

I downloaded the corresponding jar for SQL server and placed it in the shared folder which is read by Tomcat/Jetty server. So loading the SQL Server class should not be a problem according to the server docs.

PR #124, opened 2020-08-15
Update Dockerfile

jvitrifork:
Added caching of maven dependency resolving

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,27 +1,13 @@
-FROM ubuntu as build-hapi
-
-ENV PATH="/tmp/apache-maven-3.6.0/bin:${PATH}"
-
-ENV TZ=Etc/UTC
-RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
-
-WORKDIR /tmp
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,31 +1,17 @@
...
+FROM maven:3.6.3-jdk-11-slim as build-hapi
 WORKDIR /tmp/hapi-fhir-jpaserver-starter
 
-COPY . .
+COPY pom.xml .
+RUN mvn dependency:go-offline
 
-RUN /tmp/apache-maven-3.6.3/bin/mvn clean install -DskipTests
+COPY src/ /tmp/hapi-fhir-jpaserver-starter/src/
+RUN mvn clean install -DskipTests
```
seanmcilvenna, line 8: Don't you think that if the tests fail the build should fail?

jkiddo: I optimized the dockerfile - as can be seen, it ran with '-DskipTests' before. As such I haven't changed the semantics with regards to the use of maven

jkiddo: FYI @jvitrifork and @jkiddo is the same person

PR #125, opened 2020-08-17
Added spring boot example application that can be used during test ef…

jvitrifork:
…forts

Added use of spring profiles

PR #127, opened 2020-08-19
Added support for loading an IG at startup

jvitrifork:


jvitrifork:
@seanmcilvenna do you have an opinion on this feature?

src/main/resources/hapi.properties
```diff
@@ -165,3 +165,20 @@ partitioning.multitenancy.enabled=false
...
+# Fetch and install an IG tarball from an url, or
+# specify the id and version of an IG residing on
+# packages.fhir.org
+
+#implementation.guide.url=https://build.fhir.org/ig/hl7dk/dk-medcom/branches/corrections/package.tgz
+#implementation.guide.name=dk.fhir.ig.medcom-core
+#implementation.guide.version=0.8.0
+
+
+#implementation.guide.name=hl7.fhir.uv.ips
```
seanmcilvenna, line 181: Why is this here, as opposed to the guidance in the above comments?

jvitrifork: It actually loads more than one IG - it loads them transitively - all the IG's that the specified IG depends on, that is. The out commented values in the hapi.properties only serves as examples.

Loading multiple IG's would preferably require a change of format; hence https://github.com/hapifhir/hapi-fhir-jpaserver-starter/pull/125 - because you would need multivalues and maps in your configuration file (take a read on eg. https://itnext.io/why-should-we-prefer-the-yaml-file-over-the-properties-file-for-configurations-in-spring-boot-f31a273a923b for inspiration). So in essence, if accepted, I can change this PR with better multivalue support (aka. multiple IG's in a single configuration file) if https://github.com/hapifhir/hapi-fhir-jpaserver-starter/pull/125 is accepted. Until then, I wouldn't recommend any change in format to the configuration file besides the one addition in this PR.

seanmcilvenna:
Maybe we can stage merging this PR after #125 is done including a change to use YML format instead?

jkiddo:
Sure - i have just been waiting on feedback.

jvitrifork:
superseeded by https://github.com/hapifhir/hapi-fhir-jpaserver-starter/pull/125

PR #141, opened 2020-09-19
Renamed deployable war to ROOT in order to homogenize urls regardless…

jvitrifork:
… of deployment method


jvitrifork:
given that the current state is considered 'stable' - perhaps a docker image tag 5.1.1 would be appropriate?

seanmcilvenna (approved):

Issue #136, opened 2020-09-17
Tomcat app doesn't start


aggubin:
- Pulled latest code
- executed ./build-docker-image.sh
-  docker run -p 8080:8080 hapi-fhir/hapi-fhir-jpaserver-starter:latest

got the following: 

Error starting ApplicationContext. To display the conditions report re-run your application with 'debug' enabled.
2020-09-17 21:04:49.608 [main] ERROR o.s.boot.SpringApplication [SpringApplication.java:837] Application run failed
java.lang.IllegalStateException: Failed to execute ApplicationRunner
	at org.springframework.boot.SpringApplication.callRunner(SpringApplication.java:789)
	at org.springframework.boot.SpringApplication.callRunners(SpringApplication.java:776)
	at org.springframework.boot.SpringApplication.run(SpringApplication.java:322)
	at org.springframework.boot.web.servlet.support.SpringBootServletInitializer.run(SpringBootServletInitializer.java:173)
	at org.springframework.boot.web.servlet.support.SpringBootServletInitializer.createRootApplicationContext(SpringBootServletInitializer.java:153)
	at org.springframework.boot.web.servlet.support.SpringBootServletInitializer.onStartup(SpringBootServletInitializer.java:95)
	at org.springframework.web.SpringServletContainerInitializer.onStartup(SpringServletContainerInitializer.java:172)
	at org.apache.catalina.core.StandardContext.startInternal(StandardContext.java:5128)
	at org.apache.catalina.util.LifecycleBase.start(LifecycleBase.java:183)
	at org.apache.catalina.core.ContainerBase.addChildInternal(ContainerBase.java:717)
	at org.apache.catalina.core.ContainerBase.addChild(ContainerBase.java:690)
	at org.apache.catalina.core.StandardHost.addChild(StandardHost.java:705)
	at org.apache.catalina.startup.HostConfig.deployWAR(HostConfig.java:978)
	at org.apache.catalina.startup.HostConfig$DeployWar.run(HostConfig.java:1848)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at org.apache.tomcat.util.threads.InlineExecutorService.execute(InlineExecutorService.java:75)
	at java.base/java.util.concurrent.AbstractExecutorService.submit(AbstractExecutorService.java:118)
[...]

PR #144, opened 2020-09-22
Spring batch job enabled

psavva:
This PR will remove the need to run  -Dspring.batch.job.enabled=false and similar when running the hapi fhir jpa server before a few days ago.

Benefits:  Avoid changing build process, and adding command line built parameters when it's not needed

Changes, defaulted the spring batch job enabled to false in application.yaml

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Each tagged/released version of `hapi-fhir-jpaserver` is built as a Docker image
 
 ```
 docker pull hapiproject/hapi:latest
-docker run -p 8080:8080 -e "spring.batch.job.enabled=false" hapiproject/hapi:tagname
+docker run -p 8080:8080 hapiproject/hapi:tagname
 ```
 
 This will run the docker image with the default configuration, mapping port 8080 from the container to port 8080 in the host. Once running, you can access `http://localhost:8080/hapi-fhir-jpaserver/` in the browser to access the HAPI FHIR server's UI.
@@ -76,7 +76,7 @@ The easiest way to run this server entirely depends on your environment requirem
 
 ### Using jetty
 ```bash
-mvn jetty:run -Dspring.batch.job.enabled=false
+mvn jetty:run
 ```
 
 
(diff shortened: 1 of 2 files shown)


psavva:
@jamesagnew could you please review this PR as the recent changes to the implementation requires a change to the building process when it can be avoided

jamesagnew:
What is the root cause here?

We shouldn't be disabling the batch job processor, it is needed in order to run batch jobs... HAPI won't work properly without it.

jamesagnew:
@jvitrifork I assume this instruction was added when the boot stuff was added?

psavva:
The current instructions is to disable it.
Please see https://github.com/hapifhir/hapi-fhir-jpaserver-starter/blob/master/README.md#using-jetty

If it's too be enabled by default, then more parameters are missing.  Please see https://github.com/hapifhir/hapi-fhir-jpaserver-starter/issues/136

Caused by: org.springframework.batch.core.JobParametersInvalidException: There must be a valid number for readChunkSize, which is at least 1. You must include [resourceTypes] as a Job Parameter

seanmcilvenna (approved):

jamesagnew:
Why is this instruction needed though? We weren't recommending that people disable the job executor 2 weeks ago, and we make basic assumptions in hapi JPA that it is running.

Making it easier to disable the executor is a band aid that will just make people have more subtle problems.

@seanmcilvenna this should be reverted IMO. We need to address the root cause.

tadgh:
Just wanted to chime in here, this toggle, `spring.batch.job.enabled=false` is a specific boot setting to prevent ALL jobs from being run on boot, which for some reason, is SpringBoot's autoconfiguration default. https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/autoconfigure/batch/BatchAutoConfiguration.html

tadgh:
I don't think this disables batch itself, just the spring-boot magic of auto-running all configured jobs at context boot time.

seanmcilvenna:
@jamesagnew I can revert if you'd like... but I can definitely attest that I was not able to run hapi-fhir-jpaserver-starter without disabling spring.batch.job.enabled as well. The build is basically broken without this. So, I wonder if maybe it would be better to leave the PR merged, but create another PR that fixes the underlying issue _and_ removes the `spring.batch.job.enabled=false` stuff...???

jamesagnew:
Sorry - Given @tadgh it seems like I may have misunderstood what that setting was doing.

If it isn't actually disabling the batch processing, making it the default seems ok

seanmcilvenna:
Ok. I'll revert my revert. LoL

seanmcilvenna:
Ok *phew* there was nothing to revert. "master" is still good, and has this changed merged into it.

jvitrifork:
A little late to the party here I guess. I agree with @tadgh - this is a trigger on the Spring Boot AutoConfiguration part. Until the design or the proper Bean definitions as the required Batch job configurations are changed in HAPI FHIR, or added to this project, the default should be to disable **spring.batch.job** - and I agree that adding this to the yml file is the way to go (wonder why I didn't do this myself?). The thing that now makes me think, is that this setting is now *crucial* in the yml file as the server in its current state will not start up if this is set to true - just as crucial as having the right db address and so on

jvitrifork:
Perhaps a line of doc in the config file would be proper to add

Issue #137, opened 2020-09-18
How to change the server url


ZuSe:
Hi guys,

first of all thanks a lot for the amazing starter image. 
I really love the HAPI project and the simplicity of providing a ready to run docker file.

I played  around with the project during the last days and were successful to set it up on our k8s cluster.
However I was not able to change the default server_url with the new application.yaml file.

Is there any new attribute that needs to be changed?

I would love to serve the REST API on sth like  https://api.fhir.example.com/ is that possbile?

jvitrifork:
sure - thats possible. See eg. https://www.baeldung.com/tomcat-root-application or use the methods at https://github.com/hapifhir/hapi-fhir-jpaserver-starter/pull/140/files#diff-04c6e90faac2675aa89e2176d2eec7d8R103 or https://github.com/hapifhir/hapi-fhir-jpaserver-starter/pull/140/files#diff-04c6e90faac2675aa89e2176d2eec7d8R118

ZuSe:
Thanks. Which of the options is best in regards to performance and stability?
We deploy to kubernetes so the dockerfile is basically the easiest, but I don't mind build it around google distro-less if brings any great benefits over a tomcat base image.

jvitrifork:
distroless has a lower foot print and is more secure. Take a read on it ie. https://medium.com/better-programming/how-to-harden-your-containers-with-distroless-docker-images-c2abd7c71fdb.

ZuSe:
Thanks. Would you mind changing the default Dockerfile accordingly? 
We should provide people a ready to run project.
A working Dockerfile makes k8s-deployment very easy and can help client developers to spawn it quickly in any of their sandbox cloud envs

jvitrifork:
There should be no problem in making distroless the default in the Dockerfile, I would however value input for that decision from @seanmcilvenna and @jamesagnew

seanmcilvenna:
I'm not particularly familiar with "distroless" images... I get the concept based on that medium.com article, though. My chief concern is: HAPI is still very much under development, and often requires doing a little digging to figure out what's wrong. With a distroless image, the image may not contain the tools necessary to debug issues. Any thoughts on that?

jkiddo:
[...]

PR #145, opened 2020-09-24
Update Dockerfile

jvitrifork:
Proposal to go distroless as default

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,13 +5,12 @@ COPY pom.xml .
 RUN mvn -ntp dependency:go-offline
 
 COPY src/ /tmp/hapi-fhir-jpaserver-starter/src/
-RUN mvn clean install -DskipTests
+RUN mvn clean package spring-boot:repackage -Pboot
 
-FROM tomcat:9.0.38-jdk11-openjdk-slim-buster
+FROM gcr.io/distroless/java:11
 
-RUN mkdir -p /data/hapi/lucenefiles && chmod 775 /data/hapi/lucenefiles
-COPY --from=build-hapi /tmp/hapi-fhir-jpaserver-starter/target/*.war /usr/local/tomcat/webapps/
+COPY --from=build-hapi /tmp/hapi-fhir-jpaserver-starter/target/ROOT.war /app/main.war
 
 EXPOSE 8080
-
-CMD ["catalina.sh", "run"]
+WORKDIR /app
+CMD ["main.war"]


jvitrifork:
See https://github.com/hapifhir/hapi-fhir-jpaserver-starter/issues/137 for comments

jvitrifork:
@jamesagnew / @seanmcilvenna are we gonna proceed with this?

seanmcilvenna:
Generally, I'm fine with it...
But, I don't know how `CMD ["main.war"]` will work? Maybe it's just that I'm not familiar _enough_ with distroless images. Does a distroless java image know to run main.war from the cli in tomcat???

jvitrifork:
if its a repackaged spring boot application, as it is in this case, then yes. I suggest you give it a spin, try to run it if you want to see it for yourself

jvitrifork:
So are we proceeding with this?

jamesagnew:
I am not really clear on what the actual benefit is to this.. Smaller image size presumably?

We often need to ask people to take heap dumps and thread dumps, do those tools still work in this base image? Is anything else lost by changing the base image?

jvitrifork:
In that case, I suggest to default to the use of https://github.com/GoogleContainerTools/distroless#debug-images

jamesagnew:
I still don't get why this is better though. What do we gain by using this image? And do we know for sure that the JDK in the suggested image includes `jstack` and `jmap` tools?

jvitrifork:
I had no idea that there were users of this docker image that also needed to step into the container to execute eg. `jstack`. I would suspect those users would just use a local installation of tomcat or alike

jvitrifork:
The idea with distroless was to downsize and examplify a setup as close to production as possible, without being totally fit for production.

jamesagnew:
Well, I mean I'll put it like this: there are people of widely different levels of knowledge, experience, etc. who use HAPI FHIR. The reason I raise this kind of thing so often is that we do get toooons of support questions/emails/posts from people and it's hard (if not impossible) to keep up.

If someone is a beginner and reports a deadlock and I can point them to how to grab a thread dump so that we can help figure out what they are seeing, that's a good thing. I'd like to avoid having to tell them that they will need to rearchitect to not use docker.

I dunno, this is a hard balance to get right and I never feel like we get it quite right. I don't want hapi to just always strive to be the lowest common denomionator since beginners are generally not the ones building big production systems that HAPI was designed for in the first place, but I do also think that the types of people who need to squeeze every megabyte out of the docker image size for their deployment architecture are likely the ones who will be creating their own dockerfile anyway.

So with all of that said, what is your recommendation?

jvitrifork:
I get it - it's really hard to address such different needs and put it down in a box where everybody are satisfied. It more or less sounds like all options should be availble. If the user spectrum is as broad as you describe, it actually sounds like using docker way beyond what is needed here - besides the folks that just wants a FHIR server running for demo/development purposes without going into the details of what actually resides in docker. On the other hand - if people are actually skilled enough to step into a running container and execute black wizardry then distroless is probably not the right way to go. With that said, I guess we could just provide the distroless Docker recipe as part of the readme as an example and not have it as a part of the actual default docker setup.

jamesagnew:
Makes sense

jvitrifork:
Superseeded by https://github.com/hapifhir/hapi-fhir-jpaserver-starter/pull/160

PR #155, opened 2020-10-21
Docker buid - heap memory error

psavva:
When building in Docker, we sometimes face a heap memory running out when compiling the WAR file.
Setting the MAVEN_OPTS="-Xmx1024m" resolves this

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,7 @@ COPY pom.xml .
 RUN mvn -ntp dependency:go-offline
 
 COPY src/ /tmp/hapi-fhir-jpaserver-starter/src/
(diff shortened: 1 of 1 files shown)


psavva:
@jkiddo You want to check if you're ok with this change?

jkiddo:
The build runs inside docker and docker doesn't put memory constraints on itself by default so I cant see how this would help with anything

jkiddo:
See https://docs.docker.com/config/containers/resource_constraints/#:~:text=Docker%20can%20enforce%20hard%20memory,contention%20on%20the%20host%20machine. for reference

psavva:
Please revert in this case.  It must be a memory constraint in my docker build which is causing it then.  I'll investigate further

PR #171, opened 2020-11-23
Spring pr

jvitrifork:
This reflects the code changes discussed in https://github.com/hapifhir/hapi-fhir-jpaserver-starter/issues/164


src/main/resources/application.yaml
```diff
@@ -5,32 +5,31 @@ spring:
...
 
   batch:
     job:
       enabled: false
 
 hapi:
   fhir:
-    defer_indexing_for_codesystems_of_size: 101
+    ### This is the FHIR version. Choose between, DSTU2, DSTU3, R4 or R5
+    fhir_version: DSTU3
```
jamesagnew, line 31: R4 should be default, as it's the current release

jamesagnew (commented):
Looks like this still has CI issues - Overall, cool approach though!

jkiddo:
whoooops ... hadn't seen that. Standby for fix

src/main/resources/application.yaml
jvitrifork: now it is

jvitrifork:
@jamesagnew I just checked the documentation. It currently does not mention how to use the fhir_version/'spring boot active' profiles - which means that there actually isn't any documentation to change in regards to this PR - which again means that you can freely accept this PR without having 'forgotten' to update any doc as it isn't present anyway ;)

jamesagnew:
Nice, thanks for the PR!

PR #177, opened 2020-11-27
Ease setting environment configuration values

jvitrifork:


src/main/resources/application.yaml
```diff
@@ -1,6 +1,7 @@
 spring:
   datasource:
-    url: 'jdbc:h2:file:./target/database/h2'
+    #url: 'jdbc:h2:file:./target/database/h2'
+    url: jdbc:h2:mem:test_mem
```
jamesagnew, line 4: I'm not sure I love the default behaviour being a memory DB.. Feels like this will result in confused newbies.

jvitrifork: I can see what you mean. It's fixed now

jvitrifork: What about the other defaults? I've out commented the swiss one, but kept IPS as an example. Would that be ok for you?

jamesagnew: @jvitrifork IPS feels like a sensible example for sure - although probably it should be commented out by default

jkiddo: Done

jvitrifork:
Satisfied, @jamesagnew ?

jkiddo:
@jamesagnew would you mind providing me or @jvitrifork with write access?

jamesagnew (approved):

Issue #186, opened 2020-12-03
A handful of errors in config handling


jtheory:
These are things I fixed along the way in a fork of this project; these look like fixes that'd be valuable back here, too.

In AppProperties.java:
```
  // this should be false to respect the hapi-fhir default (and the default shown in application.yaml)
  private Boolean auto_create_placeholder_reference_targets = true;
...

  // in setReuse_cached_search_results_millis()... the expected value for "don't reuse" is null. Can't do that in yaml; I did this:
  public void setReuse_cached_search_results_millis(Long reuse_cached_search_results_millis) {
    if (Objects.equals(reuse_cached_search_results_millis, 0L)) {
      this.reuse_cached_search_results_millis = null;
    } else {
      this.reuse_cached_search_results_millis = reuse_cached_search_results_millis;
    }
  }
...

public static class Partitioning {

    private Boolean partitioning_include_in_search_hashes = false;
    // this was missing entirely, and the getter/setter:
    private Boolean allow_references_across_partitions = false;
```

in FhirServerConfigCommon.java
```
    // this line overrides what was set above for this property using the wrong value
    // remove (next step: figure out where allowPlaceholderReferences is supposed to go)
    retVal.setAutoCreatePlaceholderReferenceTargets(appProperties.getAllow_placeholder_references());
...

and the partitioning section did enabling, but didn't set either config value:
    // Partitioning
    if (appProperties.getPartitioning() != null) {
      retVal.setPartitioningEnabled(true);
      retVal.setIncludePartitionInSearchHashes(appProperties.getPartitioning().getPartitioning_include_in_search_hashes());
      if(appProperties.getPartitioning().getAllow_references_across_partitions()) {
        retVal.setAllowReferencesAcrossPartitions(CrossPartitionReferenceMode.ALLOWED_UNQUALIFIED);
      } else {
        retVal.setAllowReferencesAcrossPartitions(CrossPartitionReferenceMode.NOT_ALLOWED); 
      }
    }
```

in application.yaml:
```
    #partitioning:
    #  cross_partition_reference_mode: true    <= change to match the real name and default: allow_references_across_partitions: false
    #  multitenancy_enabled: true              <= remove; unused
    #  partitioning_include_in_search_hashes: true   <= change default to false (match hapi-fhir default)
...

remove: unused
#    partitioning_multitenancy_enabled:

```

I think that's it! Hope this helps.

jvitrifork:
[...]

PR #200, opened 2021-01-22
issue 186: minor fixes in partitioning config

jtheory:
See https://github.com/hapifhir/hapi-fhir-jpaserver-starter/issues/186

Tests pass (though I didn't change any), which is to say that I'm fixing things that don't yet have test coverage.
(Normally I'd add them; but it's a time committment I can't cover at the moment)

### DIFF
--- a/src/main/java/ca/uhn/fhir/jpa/starter/FhirServerConfigCommon.java
+++ b/src/main/java/ca/uhn/fhir/jpa/starter/FhirServerConfigCommon.java
@@ -5,6 +5,7 @@ import ca.uhn.fhir.jpa.binstore.DatabaseBlobBinaryStorageSvcImpl;
 import ca.uhn.fhir.jpa.binstore.IBinaryStorageSvc;
 import ca.uhn.fhir.jpa.config.HibernateDialectProvider;
 import ca.uhn.fhir.jpa.model.config.PartitionSettings;
+import ca.uhn.fhir.jpa.model.config.PartitionSettings.CrossPartitionReferenceMode;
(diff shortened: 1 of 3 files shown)


src/main/java/ca/uhn/fhir/jpa/starter/FhirServerConfigCommon.java
```diff
@@ -77,7 +78,6 @@ public DaoConfig daoConfig(AppProperties appProperties) {
     retVal.setAllowMultipleDelete(appProperties.getAllow_multiple_delete());
     retVal.setAllowExternalReferences(appProperties.getAllow_external_references());
     retVal.setExpungeEnabled(appProperties.getExpunge_enabled());
-    retVal.setAutoCreatePlaceholderReferenceTargets(appProperties.getAllow_placeholder_references());
```
jamesagnew, line 80: Was removing this line deliberate? This setting isn't related to partitioning and is still useful..

jamesagnew (changes requested):

src/main/java/ca/uhn/fhir/jpa/starter/FhirServerConfigCommon.java
jtheory: Yep! Search for `setAutoCreatePlaceholderReferenceTargets` here and you'll see what I'm fixing.

BUT (from what I recall) I couldn't figure out where `getAllow_placeholder_references` was actually meant to go, so for now that setting is dead code.

jamesagnew: Got it, thanks!

PR #208, opened 2021-02-08
Added configuration option for the ApacheProxyAddressStrategy

### LINKED ISSUE #207: Configuration option to use the ApacheProxyAddressStrategy
I've had some issues with some proxy server configuration which were resolved by using the `ApacheProxyAddressStrategy`. The implementation is over at:

https://github.com/chgl/hapi-fhir-jpaserver-starter/tree/configure-apache-proxy-strategy

It adds two new configuration options:

```yaml
hapi:
  fhir:
    ### enable to use the ApacheProxyAddressStrategy which uses X-Forwarded-* headers
    ### to determine the FHIR server address
    use_apache_address_strategy: false
    ### forces the use of the https:// protocol for the returned server address.
    ### alternatively, it may be set using the X-Forwarded-Proto header.
    use_apache_address_strategy_https: false
```

If you think it makes sense to include these options in the official jpaserver-starter, I'd be happy to create a PR!

chgl:
Closes #207

### DIFF
--- a/src/main/java/ca/uhn/fhir/jpa/starter/AppProperties.java
+++ b/src/main/java/ca/uhn/fhir/jpa/starter/AppProperties.java
@@ -66,6 +66,25 @@ public class AppProperties {
 
   private Boolean lastn_enabled = false;
 
+  private Boolean use_apache_address_strategy = false;
+  private Boolean use_apache_address_strategy_https = false;
+
+  public Boolean getUse_apache_address_strategy() {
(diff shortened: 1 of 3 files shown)


jvitrifork:
I'm ok with this. @patrick-werner do you have any comments?

patrick-werner:
> I'm ok with this. @patrick-werner do you have any comments?

looking great. Thanks for the contribution @chgl

PR #210, opened 2021-02-08
Combining Repository Validating Interceptor and IG installments

jvitrifork:
@jamesagnew I would like your comments on this way of wiring the interceptors. Is different from the current way of doing it but it is logically the same. I would say it is a bit more type safe and stream lined, but I will only continue doing so if you're ok with it


jvitrifork:
@tadgh If @jamesagnew is busy doing other stuff I would value your approval / comments on this piece

tadgh:
Heya, I can read over this by EOD hopefully

src/main/java/ca/uhn/fhir/jpa/starter/BaseJpaRestfulServer.java
```diff
@@ -97,6 +97,9 @@
   @Autowired
   ApplicationContext myApplicationContext;
 
+  @Autowired(required = false)
+  RepositoryValidationInterceptorFactoryR4 factory;
+
```
tadgh, line 102: Does it make sense for the R4 factory to be in BaseJpaREstfulServer? Seems like those repositoryValidationInterceptorFactorys could inherit some interface, and you can autowire the interface.

src/main/java/ca/uhn/fhir/jpa/starter/BaseJpaRestfulServer.java
```diff
@@ -351,28 +354,24 @@ protected void initialize() throws ServletException {
     if (appProperties.getImplementationGuides() != null) {
       Map<String, AppProperties.ImplementationGuide> guides = appProperties.getImplementationGuides();
       for (Map.Entry<String, AppProperties.ImplementationGuide> guide : guides.entrySet()) {
-        packageInstallerSvc.install(new PackageInstallationSpec()
-          .setPackageUrl(guide.getValue().getUrl())
-          .setName(guide.getValue().getName())
-          .setVersion(guide.getValue().getVersion())
-          .setInstallMode(PackageInstallationSpec.InstallModeEnum.STORE_AND_INSTALL));
+			PackageInstallOutcomeJson outcome = packageInstallerSvc.install(new PackageInstallationSpec()
```
tadgh, line 357: Doesn't seem like this variable assignment is necessary, though I may just be missing its usage.

src/main/java/ca/uhn/fhir/jpa/starter/BaseJpaRestfulServer.java
```diff
@@ -351,28 +354,24 @@ protected void initialize() throws ServletException {
...
+				.setPackageUrl(guide.getValue().getUrl())
+				.setName(guide.getValue().getName())
+				.setVersion(guide.getValue().getVersion())
+				.setInstallMode(PackageInstallationSpec.InstallModeEnum.STORE_AND_INSTALL));
+
       }
     }
 
+    if(factory != null)
+    	interceptorService.registerInterceptor(factory.buildUsingStoredStructureDefinitions());
```
tadgh, line 367: Its just a style thing, but would appreciate braces around the if, even though its only one line.

tadgh (approved):
Looks ok to me with some comments.

src/main/java/ca/uhn/fhir/jpa/starter/BaseJpaRestfulServer.java
jvitrifork: Agree - all the leg work had been done so I've introduced the interface

src/main/java/ca/uhn/fhir/jpa/starter/BaseJpaRestfulServer.java
jvitrifork: Was used in a debug session. It shouldn't have survived

src/main/java/ca/uhn/fhir/jpa/starter/BaseJpaRestfulServer.java
jvitrifork: Fixed

jvitrifork:
@tadgh any change you could get @jamesagnew to read my last message I left him at zulip?

PR #233, opened 2021-05-02
fixed property naming in readme

JavaImpala:
Just suggested an addition (accepted) to the readme with the wrong property name. Fixed now.


jamesagnew (approved):
Thanks for the contribution!

PR #236, opened 2021-05-20
Fixed indentation error in application.yaml

chgl:
There was a small indentation error in the application.yaml causing the application startup to fail

We could introduce additional tooling as part of the GitHub Workflows to lint yaml and detect these potential problems earlier (eg https://github.com/github/super-linter).

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -84,11 +84,11 @@ hapi:
 #    partitioning:
 #      allow_references_across_partitions: false
 #      partitioning_include_in_search_hashes: false
-     cors:
-       allow_Credentials: true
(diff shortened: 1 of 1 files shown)


jamesagnew:
Hi @chgl - Thanks for this!

Unfortunately it is showing a merge conflict and can't be merged as is - Do you mind to rebase it?

Also, love the idea about adding an auto-linter. If you wanted to submit a PR that would be welcome for sure.

chgl:
@jamesagnew  stragen, it looks ok from my end (except for travis): 
image

> If you wanted to submit a PR that would be welcome for sure.

Great, will do!

jamesagnew:
@chgl - Huh, that's weird. I could swear it showed conflicts before but I agree it doesn't now.

Either I was reading through my nose (most likely) or it might have been a github glitch. Either way, I'll merge now.

PR #240, opened 2021-05-23
Corrected subscription environment variables

sidharthramesh:
According to the default application.yaml, it's `hapi.fhir.subscription.resthook_enabled` (with the underscore). I've tested this with the latest docker container and it works. Setting `hapi.fhir.subscription.resthook.enabled=true` does not enable subscriptions. I spent a lot of time breaking my head over this simple variable name. Hopefully, this will help someone.


jamesagnew (approved):
Thanks for the contribution! Will merge now

PR #244, opened 2021-05-27
Changing cached search results time

luisberga:
I think is important for new user to know this feature.


jvitrifork:
@luisberga this can be controlled by the client which in most cases is a better choice, I believe: https://hapifhir.io/hapi-fhir/docs/server_jpa/configuration.html#search-result-caching

jkiddo:
I read it wrong. Yes, you are right

PR #217, opened 2021-03-13
Added GitHub actions to build container images

chgl:
This adds a GitHub Actions Workflow to automatically build images from the Dockerfile and push them to both the GHCR.io and the Docker Hub registries.

For this to work, the following repository secrets will need to be created:

- `DOCKERHUB_TOKEN`: the hub.docker.com PAT for the `hapiproject` organization
- `DOCKERHUB_USERNAME`: the username associated with the PAT above
- `GHCR_TOKEN`: a GitHub PAT used to push the image to GHCR.io (GHCR support may need to be enabled for the organization first, see https://docs.github.com/en/packages/guides/enabling-improved-container-support)

Note that the use of GHCR.io is just a suggestion to provide additional registry due to docker.io's recent rate limiting. quay.io would also be a viable choice.

The images are built on each PR, but only pushed if a new tag has been created from `master` that conforms to the format `image/v*`. This was done to be able to release new images independent of the current release tags which correspond to upstream HAPI releases.

As an example, in my fork I created the tag `v1.1.1-hapi5.3.0` (https://github.com/chgl/hapi-fhir-jpaserver-starter/releases/tag/image%2Fv1.1.1-hapi5.3.0) which triggered the workflow and created images tagged as `v1.1.1-hapi5.3.0`. See https://hub.docker.com/r/chgl/hapi-fhir-jpaserver-starter/tags?page=1&ordering=last_updated.

Note that (mostly for my personal convenience!) I've also updated the Dockerfile to make it easier to build `distroless` images and included a workflow job to build images tagged as `*-distroless`.

It's quite an opiniated first draft, so I am open to any change-requests!

Closes #205 

(this renders the "Java CI with Maven" redundant, so it could be removed)

PR #248, opened 2021-06-05
Adding dependency to snakeyaml

seanmcilvenna:
At least a few people (myself included) have had problems launching `hapi-fhir-jpaserver-starter` without having snakeyaml added as an explicit dependency.
I'd like merge this change to add it as a dependency. After it is merged, I'd like to delete the tag for 5.4.0 and re-tag it on the HEAD of `master` (including this change) as 5.4.0.
This will cause the docker image to rebuild for 5.4.0, and should solve the problem for at least a few people, if not more.
Please let me know if you're ok with me deleting the 5.4.0 tag and re-creating it on the HEAD of master when you approve/reject this merge request.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,6 +64,11 @@
             <artifactId>postgresql</artifactId>
             <version>42.2.9</version>
         </dependency>
+        <dependency>
+            <groupId>org.yaml</groupId>
+            <artifactId>snakeyaml</artifactId>
+            <version>1.28</version>
+        </dependency>
 
         <!-- Needed for Email subscriptions -->
         <dependency>


seanmcilvenna:
Or, maybe I should create a 5.4.1 tag instead? Please let me know.

jvitrifork:
it is already present with https://github.com/hapifhir/hapi-fhir-jpaserver-starter/commit/787a1b8d0f872e08c6e21c3aa139cc547564db5a
If we wan't to be stringent, I would remove the 5.4.0 and make a new from head called 5.4.1 - eventhough that also could cause confusion.

jamesagnew (approved):

jkiddo:
@jamesagnew you just duplicated the dep

jamesagnew:
whoops, lol. will revert, thanks for the heads up @jkiddo

jvitrifork:
@seanmcilvenna feel free to publish a new docker image

chgl:
It might make sense to version the image independently of the HAPI FHIR releases. For example, the people at Strimzi (https://github.com/strimzi/strimzi-kafka-operator) tag their images using the internal Strimzi version followed by the supported Kafka version, eg: a published image tag looks like `0.23.0-rc1-kafka-2.8.0`. In case of the hapi-fhir-starter, one could use `0.1.0-hapi-5.4.0`. So this bugfix would publish `0.1.1-hapi-5.4.0` - the GitHub workflow I added in https://github.com/hapifhir/hapi-fhir-jpaserver-starter/pull/217 would already allow this.

PR #255, opened 2021-06-27
Bugfix/mdm fix

jkiddo:
Made MDM registration work


jkiddo:
There is still some leftovers to be fixed in the core project, but this branch makes it possible to actually use MDM

PR #274, opened 2021-09-14
Adds ability to control the number of executor threads through environment variables.

theGOTOguy:
Currently, the HAPI FHIR JPA server has a fixed number (10) of executor threads.  It is not possible to change this default without forking the server.  We are currently unable to completely utilize the available CPU and memory in our clusters  because we are unable to tune the number of executor threads to something more appropriate.

This fixes the issue by copying the current Tomcat default server.xml and catalina.properties files into the root of the JPA server and modifying them such that:

1.  Properties in server.xml are taken from the local environment variables
2.  server.tomcat.max-threads and server.tomcat.min-spare-threads control the maximum and minimum number of executor threads available on the server by modifying the default connector in server.xml.

In the long term, this is probably a responsible thing to do because more configuration options could be exposed to users of the JPA server through the server.xml file as necessary.

This is not a breaking change and does not affect the behavior of the JPA server unless these environment variables are set.

You may test to confirm as such:

```bash
docker build .
```
(note the build hash)

```bash
docker run -p 8080:8080 hash-you-noted-above
```
See that you still have 10 executor threads, the default behavior.

```bash
docker run -p 8080:8080 -e server.tomcat.max-threads=20 -e server.tomcat.min-spare-threads=20 hash-you-noted-above
```
See that you now have 20 executor threads, thus allowing you to tune your server's parallelism.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,6 +2,7 @@ FROM maven:3.6.3-jdk-11-slim as build-hapi
 WORKDIR /tmp/hapi-fhir-jpaserver-starter
 
 COPY pom.xml .
+COPY server.xml .
 RUN mvn -ntp dependency:go-offline
 
 COPY src/ /tmp/hapi-fhir-jpaserver-starter/src/
@@ -26,6 +27,9 @@ FROM tomcat:9.0.38-jdk11-openjdk-slim-buster
 
 RUN mkdir -p /data/hapi/lucenefiles && chmod 775 /data/hapi/lucenefiles
 COPY --from=build-hapi /tmp/hapi-fhir-jpaserver-starter/target/*.war /usr/local/tomcat/webapps/
+
+COPY catalina.properties /usr/local/tomcat/conf/catalina.properties
+COPY server.xml /usr/local/tomcat/conf/server.xml
 
 EXPOSE 8080
 
--- a/catalina.properties
+++ b/catalina.properties
@@ -1,0 +1,208 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#
+# List of comma-separated packages that start with or equal this string
+# will cause a security exception to be thrown when
+# passed to checkPackageAccess unless the
+# corresponding RuntimePermission ("accessClassInPackage."+package) has
+# been granted.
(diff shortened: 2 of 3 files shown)


jvitrifork:
Should the port (8080) as well be parameterized while we're at it?

jvitrifork:
You're right on that one. I can actually be set build time (https://stackoverflow.com/questions/52591197/exposing-configurable-argument-port-at-build/52591294) but it provides little value in this setup. Also, the `expose` is 'only' documentation.

jvitrifork:
and whats the problem just specifying it in the yaml as stated on https://stackoverflow.com/questions/25399400/maximum-client-request-thread-pool-size-in-spring

Example for specifying server attributes can be found at https://docs.spring.io/spring-boot/docs/current/reference/html/howto.html#howto.properties-and-configuration.yaml

theGOTOguy:
> and whats the problem just specifying it in the yaml as stated on https://stackoverflow.com/questions/25399400/maximum-client-request-thread-pool-size-in-spring

That was actually the first thing I tried.  Both `server.tomcat.max-threads` and `server.tomcat.threads.max` are ignored, because those only work for embedded Tomcat, not for the Catalina wrapper that the JPA server uses.  It's straightforward to confirm that with:

```
docker run -p 8080:8080 -e server.tomcat.max-threads=1 hapiproject/hapi:latest
```

Send a few requests and you'll see that you still have 10.  Same with `server.tomcat.threads.max`.

theGOTOguy:
>  Also, the `expose` is 'only' documentation.

Wow, you learn something new every day!  I removed it and, you're correct.  It works just the same.

Added a commit which makes the port configurable with the environment variable `server.port`.  You can confirm that the change isn't breaking with:

```
docker build .
docker run -p 8080:8080 your_build_hash_here
```
And the server is visible at localhost:8080

But then you can change the port with:
```
docker run -p 8081:8081 -e server.port=8081 your_build_hash_here
```
And the server is now visible at localhost:8081

Some users will likely appreciate this feature.  I know there have been times when we've wished to run two separate instances of HAPI FHIR in the same Knative pod and found the inability to change port limiting.

jvitrifork:
Right - and you are ofc. right. Just checked it out and remembered that the reason why the items I referenced don't work is because we currently DON'T run in an embedded tomcat -🤦‍♂️

jvitrifork:
@tadgh / @jamesagnew / @seanmcilvenna / @patrick-werner / @oliveregger - do you guys have a feeling about the use of this project? Do people still build it and deploy it separatly to their own tomcat/whatever server? Should we keep having that option open?

jvitrifork:
It would lower the barrier for Spring Boot native people just to have it running using embedded Tomcat, but if the crowd is wider, I open to still keep it open.

tadgh:
There are plenty of users who use this and deploy it to their own tomcat servers, absolutely. Generally in those cases they will build the war and deploy it manually as opposed to using the docker container. I'm not opposed to this,  would probably be good for @jamesagnew to weigh in though

jamesagnew:
Yeah definitely agreed - For better or worse, there are still a lot of people who want to deploy a war file to a container.

Not my preferred way of standing up a server, but much like our still supporting JDK8 it's a necessity in a lot of environments

jvitrifork:
Thinking about it - I think we might should consider to have two different ways of building this project - one for Spring Boot people with Embedded Tomcat in mind - and one for building a war deployable. Right now we are in between, which might not be the best option.

jkiddo:
@theGOTOguy you should now be able to control the settings that you mention using plain environment variables if you use the distroless images - like `hapiproject/hapi:v5.7.0`

Issue #264, opened 2021-08-25
5.5.0 and Spring Boot


jvitrifork:
@jamesagnew there's some work to be done before the starter project can absorb 5.5.0. 

As can be seen on the https://github.com/hapifhir/hapi-fhir-jpaserver-starter/tree/rel_5_5_0 branch - the following is printed in the console when running the tests

```
java.lang.IllegalStateException: Failed to load ApplicationContext
Caused by: org.springframework.beans.factory.support.BeanDefinitionOverrideException: Invalid bean definition with name 'scopedTarget.reindexWriter' defined in class path resource [ca/uhn/fhir/jpa/reindex/job/ReindexEverythingJobConfig.class]: Cannot register bean definition [Root bean: class [null]; scope=step; abstract=false; lazyInit=null; autowireMode=3; dependencyCheck=0; autowireCandidate=false; primary=false; factoryBeanName=ca.uhn.fhir.jpa.reindex.job.ReindexEverythingJobConfig; factoryMethodName=reindexWriter; initMethodName=null; destroyMethodName=(inferred); defined in class path resource [ca/uhn/fhir/jpa/reindex/job/ReindexEverythingJobConfig.class]] for bean 'scopedTarget.reindexWriter': There is already [Root bean: class [null]; scope=step; abstract=false; lazyInit=null; autowireMode=3; dependencyCheck=0; autowireCandidate=false; primary=false; factoryBeanName=ca.uhn.fhir.jpa.reindex.job.ReindexJobConfig; factoryMethodName=reindexWriter; initMethodName=null; destroyMethodName=(inferred); defined in class path resource [ca/uhn/fhir/jpa/reindex/job/ReindexJobConfig.class]] bound.

```

Do you have someone on it or is it up to the community to get that fixed?

tadgh:
Ping me on November 10th as a reminder and I'll make sure i spin up before HAPI's next quarterly release 👍🏻

jvitrifork:
@tadgh PING! PING! PING! PING!

PR #291, opened 2021-11-18
5.6.0 upgrade

jkiddo:


jkiddo:
@tadgh Good to see that you got rid of all the duplicate bean definitions (as discussed on https://github.com/hapifhir/hapi-fhir-jpaserver-starter/issues/264). Next time please remember to take the project for a spin - now another property is needed as flyway needs to be disabled by default (https://github.com/hapifhir/hapi-fhir-jpaserver-starter/blob/5.6.0-upgrade/src/main/resources/application.yaml#L14)

Issue #280, opened 2021-10-05
Default to Spring Boot behaviour


jvitrifork:
As discussed on https://github.com/hapifhir/hapi-fhir-jpaserver-starter/pull/274 - it might be worth to default to regular Spring Boot behaviour in our Docker images. The project should still support that a regular WAR file can be built - but there is little reason for this to be the default choice in the Docker file which instead should default to using Spring Boot. The assumption here is that the users that need a WAR file is not the same people that uses Docker.

As far as I can see it only expands the configuration options defaulting to the use of the self-contained Spring Boot app within the container. @tadgh / @jamesagnew / @seanmcilvenna / @chgl  let me hear what you guys think?

(the functionality of running solely as Spring Boot in the container is obtained by removing the bottom 5 lines from the Dockerfile - thats it)

chgl:
Sounds good to me! Another possible enhancement would also be leveraging jar layering: https://docs.spring.io/spring-boot/docs/current/maven-plugin/reference/htmlsingle/#packaging.layers and https://spring.io/guides/topicals/spring-boot-docker/ (see "Spring Boot Layer Index"). 

The distroless "flavor" of the container image should already be using an embedded tomcat.

ZuSe:
Agree.
There is also distroless with java 17 close by. Shall we upgrade to it? It tried it out for some weeks, the whole project seems to work with it.
Only think I had to do was updating maven to 3.8.3

jkiddo:
@ZuSe will you draft a PR?

jkiddo:
@ZuSe will you still draft the PR?

jkiddo:
Default spring behavior can now be used with the distroless images. Fixed with https://github.com/hapifhir/hapi-fhir-jpaserver-starter/pull/319

PR #305, opened 2022-01-12
Upgrade/java17

ZuSe:
This closes https://github.com/hapifhir/hapi-fhir-jpaserver-starter/issues/280

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM maven:3.8.2-jdk-11-slim as build-hapi
+FROM maven:3.8-openjdk-17-slim as build-hapi
 WORKDIR /tmp/hapi-fhir-jpaserver-starter
 
 COPY pom.xml .
@@ -10,10 +10,9 @@ RUN mvn clean install -DskipTests
 
 FROM build-hapi AS build-distroless
 RUN mvn package spring-boot:repackage -Pboot
-RUN mkdir /app && \
-    cp /tmp/hapi-fhir-jpaserver-starter/target/ROOT.war /app/main.war
+RUN mkdir /app && cp /tmp/hapi-fhir-jpaserver-starter/target/ROOT.war /app/main.war
 
-FROM gcr.io/distroless/java-debian11:11 AS release-distroless
+FROM gcr.io/distroless/java17:nonroot
 COPY --chown=nonroot:nonroot --from=build-distroless /app /app
 # 65532 is the nonroot user's uid
 # used here instead of the name to allow Kubernetes to easily detect that the container
@@ -22,12 +21,13 @@ USER 65532:65532
 WORKDIR /app
 CMD ["/app/main.war"]
 
-FROM tomcat:9.0.53-jdk11-openjdk-slim-bullseye
-
-RUN mkdir -p /data/hapi/lucenefiles && chmod 775 /data/hapi/lucenefiles
-COPY --from=build-hapi /tmp/hapi-fhir-jpaserver-starter/target/*.war /usr/local/tomcat/webapps/
-
-COPY catalina.properties /usr/local/tomcat/conf/catalina.properties
-COPY server.xml /usr/local/tomcat/conf/server.xml
-
-CMD ["catalina.sh", "run"]
+# Use this for local development or deployments with shell access required
+# FROM tomcat:9.0.53-jdk11-openjdk-slim-bullseye
+#
+# RUN mkdir -p /data/hapi/lucenefiles && chmod 775 /data/hapi/lucenefiles
+# COPY --from=build-hapi /tmp/hapi-fhir-jpaserver-starter/target/*.war /usr/local/tomcat/webapps/
+#
+# COPY catalina.properties /usr/local/tomcat/conf/catalina.properties
+# COPY server.xml /usr/local/tomcat/conf/server.xml
+#
+# CMD ["catalina.sh", "run"]
--- a/pom.xml
+++ b/pom.xml
@@ -14,18 +14,18 @@
     <parent>
         <groupId>ca.uhn.hapi.fhir</groupId>
         <artifactId>hapi-fhir</artifactId>
-		  <version>5.6.0</version>
+        <version>5.6.0</version>
     </parent>
 
     <artifactId>hapi-fhir-jpaserver-starter</artifactId>
 
     <properties>
-        <java.version>8</java.version>
+        <java.version>17</java.version>
         <spring_boot_version>2.5.6</spring_boot_version>
     </properties>
 
     <prerequisites>
-        <maven>3.6.3</maven>
+        <maven>3.8.3</maven>
     </prerequisites>
 
     <packaging>war</packaging>


jkiddo:
@ZuSe I'm happy to do this - @jamesagnew / @seanmcilvenna  you usually have oppinions on the Java version due to your knowledge about who uses it. Are we good to go regarding upgrading Java?

jkiddo:
@tadgh could you ping @jamesagnew on this PR?

tadgh:
I'm sure we have a whack of people who are using this that are still stuck on java 8 for enterprise reasons. Does this MR effectively force the usage of JDK17? If so, i would hesitate to bump this right now. The HAPI-FHIR project is dropping support for jdk8 post 5.7.0, and increasing the minimum to 11. That said, i'll reach out to james and get his thoughts.

jkiddo:
@tadgh and its safe to assume that using a docker variant is not suitable for 'enterprise reasons' ... ?

jamesagnew:
We could bump the docker image to JDK 17, but we can't bump the compile version in the pom to Java 17 because that means that you can't build or use the output on anything less.

ZuSe:
@jamesagnew Why would u want to use the output on anything less? I mean the pom only counts for the JPA starter right? Java17 has been there for quite a while and is considered the current LTS release. Even Debian managed to put it to their main repositories by now, so why stick with Java 8 oder 11? Whoever is able to run the starter should be able to upgrade java to latest LTS.

jamesagnew:
Well, I wouldn't personally want to - but there are many organizations out
there that use HAPI FHIR and have policies dictating which JDK level they
can use. I don't know why really, but there are lots of them.

Part of why we released the community survey last year was to gauge
people's readiness to upgrade. Based on that, our plan is to move to JDK 11
after 5.7.0 is released, but with a promise to backport security fixes to a
JDK 8 supported version for 1 year.

We owe it to our community to give lots of notice for a change like this,
even if it's annoying.

jkiddo:
I think it is possible to set the version using maven profiles as eg. https://www.stephenenright.com/maven-java-version. Then at least the source and target can default to Java 8/11 but when using the docker image (which I guess enterprises don't have an oppinion on - because its within docker) we can compile it using Java 17.

rogtrifork:
A bit off topic but in regards to the Dockerfile changes:
1. The current state of the Dockerfile in the PR would imply that the default image would be changed to distroless and building the version with a shell would be optional, is this a good idea? People might be depending on the shell for whatever purposes given potential legacy environments so removing it outright doesn't seem right, IMO the image structure as are now should probably be replaced to:
    * **hapiproject/hapi** - Based on distroless image, would enable higher security by default
    * **hapiproject/hapi-debug** - The old hapi image with a shell included
I'd see this more as a sane default that would require minimal changes for people that rely on the shell being there (also would allow for quality-of-life for debugging purposes for projects that rely on the images present on docker-hub). Naturally, also fix the GitHub actions to reflect said change and get green checks before merging the PR.
2. As of right now, the **default** hapi image (_hapiproject/hapi:latest_) on docker-hub requires root to run as the `usr/local/tomcat` directory is owned by root thus failing with `` java.io.IOException: Unable to create the directory [/usr/local/tomcat/webapps/ROOT]`` when run as non-root, IMO this should be changed to be non-root by default given the inevitability of a pod breaking on a cloud provider due to the user being root.

jkiddo:
@rogtrifork  I get the point to some degree. I think we can all agree on that shell'ing inside containers usually are not a good idea and that is should be avoided especially in production environments. If debugging is needed I would rather like to base it from https://github.com/GoogleContainerTools/distroless#debug-images instead of the current image as the current image relys on deploying into a tomcat instance within the docker image instead of just running regular Spring Boot. Regardless, publishing two different images will also require changes to the GH action (which should be pretty doable)

jkiddo:
> Well, I wouldn't personally want to - but there are many organizations out there that use HAPI FHIR and have policies dictating which JDK level they can use. I don't know why really, but there are lots of them. Part of why we released the community survey last year was to gauge people's readiness to upgrade. Based on that, our plan is to move to JDK 11 after 5.7.0 is released, but with a promise to backport security fixes to a JDK 8 supported version for 1 year. We owe it to our community to give lots of notice for a change like this, even if it's annoying.

@jamesagnew do you consider the starter project part of that promise? Backporting to JDK8 for a year?

jamesagnew:
Yes, this repo is absolutely a part of HAPI FHIR and is covered by our policies.

jamesagnew:
And I guess to elaborate on that- This doesn't mean that the docker container has to use a lower java since that doesn't affect anyone, but there are absolutely people who check this project out, build a WAR, and deploy to enterprise environments that don't or can't run the latest and greatest JDK. That workflow has to be preserved unless there is broad community consultation that concludes it's not needed anymore.

jkiddo:
I'm just pointing out that there are apparently many people using the docker version of this project.

jamesagnew:
I have zero doubts of that fact. I know plenty of people who use the docker builds successfully. And as I said, JDK version shouldn't affect those people so by all means let's bump it in the Dockerfile. We are using the Amazon Corretto JDK 17 image as the base for Smile CDR's docker distribution these days, it has proven to be the best performing option by a fairly decent margin and personally that would get my vote for this project too, but anything JDK17 seems like a good choice.

That doesn't change the fact that not everyone uses it though. HAPI FHIR is an open community with people working on all kinds of solutions in all kinds of environments. It gets deployed everywhere from huge clusters in the cloud, to locked down work laptops in connectathons, to ancient servers running in closets in low resource settings, to mobile devices, etc. We have taken great pains now for 22+ years to make sure we choose minimum requirements that allow people to solve important problems using HAPI regardless of which of those buckets they fall into.

This is important to me and important to our community, and it is not negotiable.

jkiddo:
JDK 8 it is then

jkiddo:
@jamesagnew how about the change to default to Spring Boot instead of the WAR setup. Is that okay with you - while we also keep the WAR supported.

jkiddo:
We could do something like the following - (very similar to the content of this PR):

```
FROM maven:3.8-openjdk-17-slim as build-hapi
WORKDIR /tmp/hapi-fhir-jpaserver-starter

COPY pom.xml .
COPY server.xml .
RUN mvn -ntp dependency:go-offline

COPY src/ /tmp/hapi-fhir-jpaserver-starter/src/
RUN mvn clean install -DskipTests

FROM build-hapi AS build-distroless
RUN mvn package spring-boot:repackage -Pboot
RUN mkdir /app && cp /tmp/hapi-fhir-jpaserver-starter/target/ROOT.war /app/main.war

FROM tomcat:9.0.58-jdk11-openjdk-slim-bullseye as tomcat

RUN mkdir -p /data/hapi/lucenefiles && chmod 775 /data/hapi/lucenefiles
COPY --from=build-hapi /tmp/hapi-fhir-jpaserver-starter/target/*.war /usr/local/tomcat/webapps/

COPY catalina.properties /usr/local/tomcat/conf/catalina.properties
COPY server.xml /usr/local/tomcat/conf/server.xml

CMD ["catalina.sh", "run"]


FROM gcr.io/distroless/java17:nonroot as default
COPY --chown=nonroot:nonroot --from=build-distroless /app /app
# 65532 is the nonroot user's uid
# used here instead of the name to allow Kubernetes to easily detect that the container
# is running as a non-root (uid != 0) user.
USER 65532:65532
WORKDIR /app
CMD ["/app/main.war"]

```

From the action scripts we can then do eg. `docker build --target tomcat . -t tomcat` which then builds the docker image where the app is deployed into tomcat as usual, but the default image will be distroless. That doesn't adress @rogtrifork 's finding though - the one about tomcat running as root. I guess we can look into https://hub.docker.com/r/bitnami/tomcat/ for the root part if needed

jkiddo:
Comments anyone?

jkiddo:
Superseeded by https://github.com/hapifhir/hapi-fhir-jpaserver-starter/pull/319

Issue #318, opened 2022-02-21
Unable to start using Docker Compose and MySQL database


MacL3an:
I'm just trying to follow the instructions in the readme, starting the HAPI Server using the Docker compose file running MySQL. I've updated the file `src/main/resources/application.yaml` to contain the following:

```
  datasource:
    url: 'jdbc:mysql://hapi-fhir-mysql:3306/hapi'
    username: admin
    password: admin
    driverClassName: com.mysql.jdbc.Driver
    max-active: 15
```

And I then run:
`docker-compose up -d --build`

I have attached the logs. Some potential clues include:
```
2022-02-21 14:09:23.616 [main] WARN  c.z.hikari.util.DriverDataSource [DriverDataSource.java:68] Registered driver with driverClassName=com.mysql.jdbc.Driver was not found, trying direct instantiation.
...
[...]

PR #321, opened 2022-02-22
Added instructions about removing Hibernate dialect

MacL3an:
To fix the problem raised in [this](https://github.com/hapifhir/hapi-fhir-jpaserver-starter/issues/318) issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,13 @@ spring:
     password: admin
     driverClassName: com.mysql.jdbc.Driver
 ```
+
+Also, make sure you are not setting the Hibernate dialect explicitly, in other words remove any lines similar to:
+
+```
+hibernate.dialect: {some none MySQL dialect}
+```
+
 On some systems, it might be necessary to override hibernate's default naming strategy. The naming strategy must be set using spring.jpa.hibernate.physical_naming_strategy. 
 
 ```yaml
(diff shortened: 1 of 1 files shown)


jkiddo:
@MacL3an your suggested change contradicts https://github.com/hapifhir/hapi-fhir-jpaserver-starter/blob/deefd595dd0189efe9f1945a7aac7757a04f3cba/src/main/resources/application.yaml#L23

patrick-werner:
I think this can be closed, application.yaml has documentation on when to add dialect, and when not to.
@MacL3an @jkiddo

MacL3an:
> #Hibernate dialect is automatically detected except Postgres and H2. 

Yes, but since the default value in the application.yaml file seems to be H2, I thought it would be nice a courtesy to users to remind them that they have to manually remove this default?

> I think this can be closed, application.yaml has documentation on when to add dialect, and when not to.

There are some notes, but since the Readme actually has a explicit section on what to do when running MySQL, why not make that section complete? Right now it only contains some of the steps needed to run MySQL, not of all of them. Instead people have to read through the application.yaml file as well, understand it and make changes that are not mentioned in the readme before they can run HAPI with MySQL.

jkiddo:
@MacL3an I guess you have a point there

patrick-werner (approved):

patrick-werner:
> @MacL3an I guess you have a point there

i agree

PR #324, opened 2022-03-03
add dao_scheduling_enabled configuration property

jschneid-nmdp:
This PR adds a dao_scheduling_enabled configuration property, to make the DaoConfig scheduling setting configurable.

In a clustered environment, DaoConfig scheduling should be disabled on all nodes except one.

The following message thread contains a few comments from James Agnew on the topic.

https://groups.google.com/g/hapi-fhir/c/kjY-o7hMRuo/m/ShNI3xl3AwAJ

"There are many scheduled maintenance jobs that must be executed once and only once per scheduling period. The scheduler must be disabled on all nodes except one."

"The DaoConfig one is the one that should be disabled."


jkiddo:
@jschneid-nmdp thx!

PR #337, opened 2022-03-20
Add actuator

jkiddo:


XcrigX:
I would recommend not committing the change to application.yaml which opens all the actuator endpoints. This could open somebody up to security risks if they take the latest update without turning those explicitly back off. 
/logfile, /heapdump, /configprops, etc. could all expose sensitive information. 

Perhaps just explicitly enabling /health - which is the only endpoint Spring exposes by default would be better, so users can see that actuator is an option, but without turning them all on. 
https://docs.spring.io/spring-boot/docs/current/reference/html/actuator.html#actuator.endpoints.security

This is the part I'm talking about:  
`#Adds the option to go to eg. http://localhost:8080/actuator/env for seeing the running configuration
management:
  endpoints:
    web:
      exposure:
        include: "*"
        exclude: "beans"`

jkiddo:
@XcrigX sure - thats a fair reasoning. Will you add a PR to the branch?

XcrigX:
Certainly.. will submit it shortly

XcrigX:
That was harder than I thought.. turns out github wont let you have a fork and a fork of a fork for the same project.. 
submitted: https://github.com/hapifhir/hapi-fhir-jpaserver-starter/pull/338

patrick-werner (approved):

PR #343, opened 2022-04-06
add support for ms sql server

XcrigX:
MS SQL Server drivers are now publicly available in Maven Central and open-sourced under MIT license. 
see: https://github.com/microsoft/mssql-jdbc

If there were not other reasons for excluding it, request to add the dependency to make it easier to use directly with SQL Server without needing to manually add the driver jar files to deployments. 
Also added a note to the readme to save people some headaches when using MS SQL Server.

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,6 +215,26 @@ spring:
 
 Because the integration tests within the project rely on the default H2 database configuration, it is important to either explicity skip the integration tests during the build process, i.e., `mvn install -DskipTests`, or delete the tests altogether. Failure to skip or delete the tests once you've configured PostgreSQL for the datasource.driver, datasource.url, and hibernate.dialect as outlined above will result in build errors and compilation failure.
 
+### Microsoft SQL Server configuration
+
+To configure the starter app to use MS SQL Server, instead of the default H2, update the application.yaml file to have the following:
+
+```yaml
+spring:
+  datasource:
+    url: 'jdbc:sqlserver://<server>:<port>;databaseName=<databasename>'
+    username: admin
+    password: admin
+    driverClassName: com.microsoft.sqlserver.jdbc.SQLServerDriver
+```
+
+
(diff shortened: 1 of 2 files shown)


jkiddo:
The HikariCP stuff needs to be changed upstream in the https://github.com/hapifhir/hapi-fhir project

XcrigX:
@jkiddo , can you elaborate regarding HikariCP?  I haven't noticed any problems running against SQL Server thus far. Thanks.

jkiddo:
You excluded it ( as a transitive dependency) and changed it explicitly. It should be fixed in the core project, preferably

jamesagnew:
FWIW it looks like this comes in as a transitive dependency from Quartz. Sigh, looks like Quartz brings in both HikariCP-java7 and c3p0. Both of these things are connection pools and neither is actually mandatory by Quartz (all it needs is a datasource instance). Someone over there clearly doesn't understand maven.

image

jamesagnew:
Hmm, looks like Quartz even recommend excluding that dependency. https://github.com/quartz-scheduler/quartz/wiki/How-to-Use-DB-Connection-Pool

jkiddo:
Oh my ... @jamesagnew I'll leave it to you to recommend how we should proceed.

XcrigX:
Ah. If I remember correctly it was the Spring Boot Actuator import that has a problem w the Hikari version brought in by Quartz.. I'm away from my computer today to look closer. Related thought, would it make things easier to import the SpringBoot bom in the upstream hapi parent and let spring manage (most of) the dependency versions?

XcrigX:
I need to look at this PR. I didn't intend to include the Actuator here and the related Hikari exclusion. That was an earlier PR which perhaps isn't merged to master yet.
I must have made a mistake. I can't troubleshoot til I get back to my computer though. If you reject this I can sort it out and resubmit w just the SQL server commit.
The Hikari problem still remains the PR related to SB Actuator though..

XcrigX:
Resubmitted a clean PR to replace this - #347

jamesagnew:
FYI I'm dropping in the hikari and c3p0 deps in https://github.com/hapifhir/hapi-fhir/pull/3528

PR #345, opened 2022-04-07
Update application.yaml

jkiddo:


jkiddo:
This PR makes MDM not activated by default

PR #355, opened 2022-04-25
Expose Prometheus metrics

chgl:
This enables the Prometheus actuator endpoint to expose application metrics on `/actuator/prometheus` (https://docs.spring.io/spring-boot/docs/current/reference/html/actuator.html#actuator.metrics.export.prometheus).

This endpoint is read-only and included by default. I don't believe any critical information is exposed via this endpoint, but we could also run the actuator on different port (via the `management.server.port` property) so it can be firewalled off if desired, combined with `management.endpoint.health.probes.add-additional-paths=true`, health probes can still run on the main server port (see also https://docs.spring.io/spring-boot/docs/current/reference/html/actuator.html#actuator.endpoints.kubernetes-probes).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -317,6 +317,13 @@
             <version>${spring_boot_version}</version>
         </dependency>
 
+        <!-- https://mvnrepository.com/artifact/io.micrometer/micrometer-registry-prometheus -->
+        <dependency>
+            <groupId>io.micrometer</groupId>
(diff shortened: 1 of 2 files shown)


madduci:
Hi, unfortunately this endpoint does not work, i get consistently the error

`[http-nio-8080-exec-1] WARN  o.s.web.servlet.PageNotFound [DefaultHandlerExceptionResolver.java:461] No endpoint GET /actuator/prometheus.`

Issue #359, opened 2022-05-04
Helm deploy with external database is failing due to missing Values.postgresql.image.registry


gsake:
Based on the instruction in values.yaml the FHIR deployment can be completed either by deploying a postgreSQL database container or using an external database by providing the host, username, password and database name in the externalDatabase section of the values.yaml file.

```
postgresql:
  # -- enable an included PostgreSQL DB.
  # see <https://github.com/bitnami/charts/tree/master/bitnami/postgresql> for details
  # if set to `false`, the values under `externalDatabase` are used
  enabled: [ false | true ]
```

Even after setting the postgresql.enabled value to false, the deployment is still trying to deploy the postegreSQL container (as it is defined in the initContainers section of the deployment.yaml template) and installation fails with the following error:

```
$ helm install --namespace test hapi-fhir-jpaserver . --values values.yaml --wait --debug

install.go:178: [debug] Original chart version: ""
install.go:199: [debug] CHART PATH: /home/gsake/repos/hapi-fhir-jpaserver-starter/charts/hapi-fhir-jpaserver

Error: INSTALLATION FAILED: template: hapi-fhir-jpaserver/templates/deployment.yaml:36:28: executing "hapi-fhir-jpaserver/templates/deployment.yaml" at <.Values.postgresql.image.registry>: nil pointer evaluating interface {}.registry
helm.go:88: [debug] template: hapi-fhir-jpaserver/templates/deployment.yaml:36:28: executing "hapi-fhir-jpaserver/templates/deployment.yaml" at <.Values.postgresql.image.registry>: nil pointer evaluating interface {}.registry
INSTALLATION FAILED
main.newInstallCmd.func2
        helm.sh/helm/v3/cmd/helm/install.go:127
github.com/spf13/cobra.(*Command).execute
        github.com/spf13/cobra@v1.2.1/command.go:856
github.com/spf13/cobra.(*Command).ExecuteC
        github.com/spf13/cobra@v1.2.1/command.go:974
github.com/spf13/cobra.(*Command).Execute
        github.com/spf13/cobra@v1.2.1/command.go:902
main.main
        helm.sh/helm/v3/cmd/helm/helm.go:87
runtime.main
        runtime/proc.go:225
runtime.goexit
        runtime/asm_amd64.s:1371
```

Although that there is a conditional dependency to postgresql in Chart.yaml, there is not a conditional postgresql installation in deployment.yaml template

```
dependencies:
[...]

PR #366, opened 2022-05-10
Conditionaly call the initContainers only when .postgresql.enable is true

gsake:
Issue #359: Helm deploy with external database is failing due to missing Values.postgresql.image.registry

Solution: The proposed solution that is verified in my environent is to conditionaly call the initContainers only when the .postgresql.enabled is set to true and to ommit if the .postgresql.enabled is set to false


jkiddo:
@chgl are you good with this?

chgl:
I commented on https://github.com/hapifhir/hapi-fhir-jpaserver-starter/pull/367 - which does the same but targets the master branch which will then trigger a chart release an will run the CI

PR #396, opened 2022-07-04
Applying fix from upstream for h2 Binaries

patrick-werner:
This issue is already fixed upstream. Can be removed with next hapi version upgrade.
https://github.com/hapifhir/hapi-fhir/pull/3676


Daanoz:
@patrick-werner I would love to have this change available in the docker version. Could you comment what the expected timeline is for a next release to hub.docker.com?

PR #400, opened 2022-07-07
fixed directory of the .war in tomcat-based image

### LINKED ISSUE #392: Tomcat-flavored container is not starting the server
When testing https://github.com/hapifhir/hapi-fhir-jpaserver-starter/pull/391 I noticed that the HAPI FHIR server wouldn't start when using the tomcat "flavored" image:

```sh
docker build -t docker.io/hapiproject/hapi:latest-tomcat --target=tomcat .
docker run --rm -it -p8080:8080 docker.io/hapiproject/hapi:latest-tomcat
```

This does start the tomcat server but HAPI itself is never deployed.

I believe the issue is in https://github.com/hapifhir/hapi-fhir-jpaserver-starter/blob/master/Dockerfile#L31 - the .war file is copied to  `/opt/bitnami/tomcat/webapps_default` instead of `/opt/bitnami/tomcat/webapps`. Changing it to the latter correctly starts the FHIR server.

If this makes sense I'd be happy to create a PR changing the .war location.

chgl:
Closes #392

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,6 @@ RUN mkdir /app && cp /tmp/hapi-fhir-jpaserver-starter/target/ROOT.war /app/main.
 FROM bitnami/tomcat:9.0 as tomcat
 
 RUN rm -rf /opt/bitnami/tomcat/webapps/ROOT && \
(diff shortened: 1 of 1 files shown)


ZuSe:
Looks good for me. Ran into the same some weeks ago!

PR #408, opened 2022-08-04
Tracking Hapi 6.2-PRE releases

jmarchionatto:
This is a tracking branch to test out work in the Hapi 6.2 snapshot stream.  The main branch is used to build against the most-recent released version of Hapi.  We will be bumping the Hapi version in pom.xml to track changes in Hapi PRE releases.

Don't merge this to master until Hapi 6.2 is released.

The main value to keep this open as a "Draft" is for for regular builds as we update the PRE release version.


tadgh:
This has now been updated to the most recent 6.2.0 PRE release. @jkiddo do you want to have a read over it? I imagine the only changes between now and the actual cut release will be the final version bump.

jkiddo:
Sure - Ill take a look

src/main/resources/application.yaml
```diff
@@ -39,16 +39,17 @@ spring:
...
+      hibernate.search.enabled: true
+  ### lucene parameters
+#      hibernate.search.backend.type: lucene
+#      hibernate.search.backend.analysis.configurer: ca.uhn.fhir.jpa.search.HapiHSearchAnalysisConfigurers$HapiLuceneAnalysisConfigurer
+#      hibernate.search.backend.directory.type: local-filesystem
+#      hibernate.search.backend.directory.root: target/lucenefiles
+#      hibernate.search.backend.lucene_version: lucene_current
+  ### elastic parameters ===> see also elasticsearch section below <===
+      hibernate.search.backend.type: elasticsearch
+      hibernate.search.backend.analysis.configurer: ca.uhn.fhir.jpa.search.HapiHSearchAnalysisConfigurers$HapiElasticAnalysisConfigurer
```
jkiddo, line 52: These lines should be disabled pr. default

src/test/resources/application.yaml
```diff
@@ -0,0 +1,186 @@
...
+#  debug:
+#    pretty_print_json_log: false
+#    refresh_after_write: false
+#  enabled: false
+#  password: SomePassword
+#  required_index_status: YELLOW
+#  rest_url: 'localhost:9200'
+#  protocol: 'http'
+#  schema_management_strategy: CREATE
+#  username: SomeUsername
```
jkiddo, line 186: I think we should have this converted to regular spring boot ElasticsearchRestClientAutoConfiguration but that is for another time to do so

jkiddo (approved):

src/main/resources/application.yaml
jkiddo: @tadgh besides the above it LGTM

tadgh: ah, cheers, will fix.

PR #422, opened 2022-09-03
Feat/restructuring

jkiddo:
This PR moves a lot around on the classes and puts the construction of the FHIR server into a separate bean for making it easier for other people to reuse it. All field injection has also been removed and put into constructors or bean creatino instead. There is still much more work to be done as with this redesign as eg. all the inceptors can be retrofitted later in construction process. This step however is the first in getting a more cleaner setup and the first step for us to be able to do more auto configuration more cleanly. Let me know what you think @tadgh @michaelabuckley @patrick-werner


jkiddo:
Ideally all the optional arguments can also be removed over time but before I go into such endeavours I would like to hear what you think of it so far.

src/main/java/ca/uhn/fhir/jpa/starter/Application.java
```diff
@@ -56,11 +57,10 @@ protected SpringApplicationBuilder configure(
...
   @Bean
   @Conditional(OnEitherVersion.class)
-  public ServletRegistrationBean hapiServletRegistration() {
+  public ServletRegistrationBean hapiServletRegistration(RestfulServer restfulServer) {
     ServletRegistrationBean servletRegistrationBean = new ServletRegistrationBean();
-    JpaRestfulServer jpaRestfulServer = new JpaRestfulServer();
-    beanFactory.autowireBean(jpaRestfulServer);
-    servletRegistrationBean.setServlet(jpaRestfulServer);
+    beanFactory.autowireBean(restfulServer);
+    servletRegistrationBean.setServlet(restfulServer);
```
tadgh, line 63: Nice, I like this refactor already

tadgh (approved):
Approved pending one question

src/main/resources/application.yaml
```diff
@@ -89,7 +89,7 @@ hapi:
     #    default_pretty_print: true
     #    default_page_size: 20
     #    delete_expunge_enabled: true
-    #    enable_repository_validating_interceptor: false
+    enable_repository_validating_interceptor: true
```
tadgh, line 92: Question (non-blocking): Did you intend to set this to true during a refactor? Seems out of place

tadgh (approved):

oliveregger (approved):

src/main/resources/application.yaml
patrick-werner: agree, this should be false by default

patrick-werner (approved):

src/main/resources/application.yaml
jkiddo: That was a bug - only enabled for testing

PR #426, opened 2022-09-12
Extract IG loading

jkiddo:
This PR extracts the installation of IG's to its own bean.


jkiddo:
This PR bringe focus on extracting IG installation. Its a bit troublesome to do as there is already a default IG installer service defined as bean - it is however insufficient in this setup as IGs with search params defined triggers reindexing eagerly.

tadgh (approved):
Looks good! I really like the ConditionalOn

jkiddo:
@tadgh any toughts on the strategy of conditionally overriding existing beans defined in the core lib? I think it brings a bit too much complexity into it, but I don’t see another way out of it.

PR #440, opened 2022-10-06
removed duplicated and wrong subscription code

### LINKED ISSUE #439: Rest Hook Subscription is always on
Due to the code: https://github.com/hapifhir/hapi-fhir-jpaserver-starter/blob/531d2557820b5ecbe725a9142dd231413a6dd716/src/main/java/ca/uhn/fhir/jpa/starter/common/FhirServerConfigCommon.java#L206

Subscriptions will be added, even if disabled.

patrick-werner:
fixes #439


patrick-werner:
you are right. Let me double check..

PR #455, opened 2022-11-12
fixes for support of R4B / 6.2.0

jkiddo:
See also https://github.com/hapifhir/hapi-fhir/issues/4276

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -14,7 +14,7 @@
     <parent>
         <groupId>ca.uhn.hapi.fhir</groupId>
         <artifactId>hapi-fhir</artifactId>
-        <version>6.2.0-PRE18-SNAPSHOT</version>
+        <version>6.2.0</version>
     </parent>
 
     <artifactId>hapi-fhir-jpaserver-starter</artifactId>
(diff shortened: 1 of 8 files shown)


jkiddo:
I've deliberately removed the subscription tests on R4B. They require more knowhow about how subscriptions work in R4B than I currently have. I encourage anyone to add that set of test cases

tadgh:
Heh you beat me to it. My task for monday was going to be to update the release branch for the new 6.2.0. did the release branch already get merged?

jkiddo:
I merged the rel_6_2_0 (I might have been a bit too quick - it wasn't listed as in 'draft') since https://github.com/hapifhir/hapi-fhir/releases/tag/v6.2.0 was released and then I thought it might make sense to go ahead.

tadgh:
Ah I see! I appreciate you taking this on :) Reviewing now

jkiddo:
(Apparently 6.2.0 is available on the snapshot repos but not on the stable repo - I imagine that the binarys will be the same?)

tadgh (approved):

jkiddo:
@tadgh then you can fix the issue I reported above on monday instead ;)

jamesagnew:
> I've deliberately removed the subscription tests on R4B

That makes sense - FYI I don't expect subscriptions would work in R4B, along with Validation.

We need to rewrite subscriptions for R5 and hopefully will get R4B in along with that... Validation is kind of "wait and see if there is demand"

vadi2:
FTR there is talk of backporting subscriptions back to R4 (not just R4B): https://chat.fhir.org/#narrow/stream/179240-Announcements/topic/R4.20Subscriptions.20Backport.20-.20public.20review

PR #456, opened 2022-11-12
Feature/static files

jkiddo:


src/main/resources/application.yaml
```diff
@@ -52,6 +52,7 @@ spring:
 #      hibernate.search.backend.analysis.configurer: ca.uhn.fhir.jpa.search.HapiHSearchAnalysisConfigurers$HapiElasticAnalysisConfigurer
 hapi:
   fhir:
+    staticLocation: file:/Users/jkv/Downloads/matchbox-main/form/
```
jamesagnew, line 55: This should probably be commented out by default with a comment about what it does

jamesagnew (approved):

PR #465, opened 2022-12-19
Update pom.xml to reference hapi-fhir 6.3.4-SNAPSHOT.  This hapi-fhir release contains a fix for POSTing an XML resource with comments results in a fhir_comments error

lukedegruchy:
- hapi-fhir 6.3.4-SNAPSHOT contains a fix for a fhir-comments error when POSTing XML resources with comments
- Upgrading required pointing to websocket-jetty dependencies that had been renamed since 6.2.2, for example 

TODO:  create a new issue to cover this

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -14,7 +14,7 @@
     <parent>
         <groupId>ca.uhn.hapi.fhir</groupId>
         <artifactId>hapi-fhir</artifactId>
-        <version>6.2.2</version>
+        <version>6.3.4-SNAPSHOT</version>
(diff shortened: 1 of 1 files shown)


jkiddo:
@lukedegruchy - the StaticLogger errors can be avoided if you do `System.setProperty("org.springframework.boot.logging.LoggingSystem","none");` - which also enables to you bump to latest available version of logback (1.4.5)

jkiddo (approved):

PR #467, opened 2022-12-21
allow interceptors to be registered via properties

XcrigX:
The purpose of this PR is to allow users to add interceptors without needing to update the existing configuration classes in HAPI - to make it easier to maintain custom branches without code-clashes in the base project. 
It adds two properties:
hapi.fhir.custom-bean-packages - any packages listed here will be ComponentScan'ed by Spring
hapi.fhir.custom-interceptor-classes - classes here will be discovered by the config (from the application context or via reflection) and registered as interceptors with the server. 

This paradigm could likely be extended for custom providers and possibly other things. 
see also : https://chat.fhir.org/#narrow/stream/179167-hapi/topic/best.20practice.20for.20embedding.20business.20logic

### DIFF
--- a/src/main/java/ca/uhn/fhir/jpa/starter/common/StarterJpaConfig.java
+++ b/src/main/java/ca/uhn/fhir/jpa/starter/common/StarterJpaConfig.java
@@ -63,9 +63,11 @@ import ca.uhn.fhir.validation.ResultSeverityEnum;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
 import org.hl7.fhir.common.hapi.validation.support.CachingValidationSupport;
+import org.springframework.beans.factory.NoSuchBeanDefinitionException;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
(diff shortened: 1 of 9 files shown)


XcrigX:
note: I somewhat combined the two approaches discussed in the zulip thread. I added _hapi.fhir.custom-bean-packages_ property to allow configurable ComponentScanning assuming that some interceptors may need to have other beans from the App Context autowired and available.  But in order to register an interceptor you still must explicitly list the interceptors in the property  _hapi.fhir.custom-interceptor-classes_.

src/main/java/ca/uhn/fhir/jpa/starter/common/StarterJpaConfig.java
```diff
@@ -75,13 +77,16 @@
...
 import javax.persistence.EntityManagerFactory;
 import javax.sql.DataSource;
+
 import java.util.*;
 
 import static ca.uhn.fhir.jpa.starter.common.validation.IRepositoryValidationInterceptorFactory.ENABLE_REPOSITORY_VALIDATING_INTERCEPTOR;
 
 @Configuration
+//allow users to configure custom packages to scan for additional beans
+@ComponentScan(basePackages = { "${hapi.fhir.custom-bean-packages:}" })
```
jamesagnew, line 87: It has never occurred to me that you could put an SpEL expression in this annotation. Neat!

jamesagnew (approved):

PR #468, opened 2022-12-21
Initial Porting of the $summary operation for IPS

psavva:
This PR serves to port the work done by  https://github.com/hapifhir/hapi-fhir/compare/master...CrossroadsLabs:hapi-fhir:rb_ips on the hapi-fhir-jpaserver-base to the hapi-fhir-jpaserver-starter

There is a lot of work needed to ensure the summary operation works as per the given implementation guide:
https://build.fhir.org/ig/HL7/fhir-ips/

Contributions to this work in welcomed.

The aim is to have this PR merged prior to the HL7 Connectathon in Las Vegas in January 2023.
Please find the info here:
https://confluence.hl7.org/display/FHIR/2023+-+01+Connectathon+32

### DIFF
--- a/.run/Application.run.xml
+++ b/.run/Application.run.xml
@@ -1,0 +1,18 @@
+<component name="ProjectRunConfigurationManager">
+  <configuration default="false" name="Application" type="SpringBootApplicationConfigurationType" factoryName="Spring Boot" nameIsGenerated="true">
+    <option name="ACTIVE_PROFILES" value="boot" />
+    <option name="ALTERNATIVE_JRE_PATH" value="BUNDLED" />
+    <option name="DEBUG_MODE" value="true" />
+    <module name="hapi-fhir-jpaserver-starter.main" />
+    <option name="SHORTEN_COMMAND_LINE" value="ARGS_FILE" />
+    <option name="SPRING_BOOT_MAIN_CLASS" value="ca.uhn.fhir.jpa.starter.Application" />
+    <extension name="software.aws.toolkits.jetbrains.core.execution.JavaAwsConnectionExtension">
+      <option name="credential" />
+      <option name="region" />
+      <option name="useCurrentConnection" value="false" />
+    </extension>
+    <method v="2">
+      <option name="Make" enabled="true" />
+    </method>
+  </configuration>
+</component>
--- a/src/main/java/ca/uhn/fhir/jpa/ig/ips/Ips.java
+++ b/src/main/java/ca/uhn/fhir/jpa/ig/ips/Ips.java
@@ -62,7 +62,6 @@ import org.springframework.context.annotation.Configuration;
  * #L%
  */
 
-@Configuration
 public class Ips {
 	private final FhirContext fhirContext;
 	private final IFhirResourceDao<Bundle> bundleResourceProvider;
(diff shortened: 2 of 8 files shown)


jddamore:
I've noted at least one file that I think will be necessary to make the narrative generation work for the Composition. See https://github.com/CrossroadsLabs/hapi-fhir/blob/rb_ips/hapi-fhir-base/src/main/resources/ca/uhn/fhir/narrative/ips/IPS.html

There may be more and I will work to review more thoroughly in the coming weeks before January 2023 connectathon.

jamesagnew:
This is really neat - I don't know that it makes sense to merge into this specific repo though. This repo is primarily concerned with packaging and configuration of the HAPI FHIR JPA server, the business logic is all in the [HAPI FHIR](github.com/hapifhir/hapi-fhir/) repo.

I'm happy to help with getting it integrated and merged over there though - I'll be at the LV Connectathon and I'm planning on participating in the IPS track as well.

psavva:
> 
Dear James, 

This was the initial take from https://github.com/hapifhir/hapi-fhir/compare/master...CrossroadsLabs:hapi-fhir:rb_ips as you are describing to put it in the  https://github.com/hapifhir/hapi-fhir repo.

However, the issues we face with this is that the $summary operation would be somewhat different for each implementation as the profiling is somewhat different for  each implementation, hence we need to have the customized mappings for each.

I'm sure we could find a way to have it merged into hapi base + custom logic for mappings could be done in the starter?

Looking forward to meeting you in Person in Vegas with the IPS track.

Best Regard
Panayiotis Savva

jamesagnew:
Sounds great! Perhaps we could spend some time at the connectathon figuring out what would be the ideal approach for getting foundational stuff into hapi-fhir, and the custom mappings into starter.

I've been pondering for a while now whether some kind of DSL might be possible that would allow people to template document generation using data from a repo - IPS being an obvious use case for it, but not the only one. Would be cool to try and plan for that being an option in the future too.

riobennin:
@psavva @jamesagnew I mentioned this on chat.fhir.org but for completeness I'll mention it here too. Our intended first approach to customization would have been to allow a custom filter and a custom narrative generator. We thought those options would be what most people would want. To implement we had discussed adding two methods to our `PatientSummary` class: `getNarrativeGenerator` and `getFilterer`. By default they would return the reference filterer and generator, but a config setting would allow them to be return custom implementations. They might implement control flow logic that says something like `if someConfig.useCustomFilterer == true` then load a custom filter class defined at an expected (or a config specified) location.

jkiddo:
> This is really neat - I don't know that it makes sense to merge into this specific repo though. This repo is primarily concerned with packaging and configuration of the HAPI FHIR JPA server, the business logic is all in the [HAPI FHIR](github.com/hapifhir/hapi-fhir/) repo.
> 
> I'm happy to help with getting it integrated and merged over there though - I'll be at the LV Connectathon and I'm planning on participating in the IPS track as well.

@jamesagnew I think I suggested to add it to the starter initially to have mature a bit and keep the complexity down for contributors. It sounds great if it can move into the core project over time. It could certainly mean a boost of the IPS standard!

jddamore:
@jkiddo We actually had @jamesagnew @riobennin and @psavva on a call this morning. Getting this all worked out and merged back into HAPI is a key goal for the connectathon so will welcome your help in making that happen!

psavva:
Superseded by https://github.com/hapifhir/hapi-fhir-jpaserver-starter/pull/487

PR #474, opened 2023-01-10
Remove old junit4 usage.

inspector-cochise:
Testcontainers sadly needs junit4. Thus some junit4 api usage appears to have crept in.

This pull request contains a tiny clean up, removing those few calls.

P.S.: I tried excluding junit4 in the pom but that does not work because testcontainers actually needs junit4 to work.

Hope this is helpful.


KevinDougan-SmileCDR (approved):
Thanks for the contribution!

PR #475, opened 2023-01-10
Feature/index default

jkiddo:
Updated docs and added redirect when only hitting part of the correct url


jkiddo:
This fix makes it easier and more 'web' like when adding eg. IG's or other static web pages where you would expect an automatic redirect to the index.html file

jamesagnew (approved):

PR #495, opened 2023-02-08
removed mysql-connector-java dependency since no longer supported and…

XcrigX:
… due to CVE-2022-3171 in transitive dependency on protobuf-java-3.19.4

MySQL is no longer supported in HAPI, and it has a transitive dependency on protobuf-java-3.19.4 which has an open CVE. 

NOTE: not certain if this is the same reason for this flag in the base HAPI project: 
https://github.com/hapifhir/hapi-fhir/issues/4373

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -55,10 +55,6 @@
             <version>${jetty_version}</version>
         </dependency>
         <dependency>
-            <groupId>mysql</groupId>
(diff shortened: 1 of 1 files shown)


jamesagnew (approved):
Thanks for doing that!

PR #499, opened 2023-02-16
Samatar/history count mode config

sabukar-fn:


src/main/resources/application.yaml
```diff
@@ -131,6 +131,11 @@ hapi:
       # These are allowed_origin patterns, see: https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/cors/CorsConfiguration.html#setAllowedOriginPatterns-java.util.List-
       allowed_origin:
         - '*'
+    # history_count_mode is an integer that represents the mode that the DAOConfig will use when getting the history count during a history request
+      # history_count_mode = 0 -> CACHED_ONLY_WITHOUT_OFFSET (default within hapi-fhir DAOConfig)
+      # history_count_mode = 1 -> COUNT_ACCURATE (ignores cache and always gets correct count)
+      # history_count_mode = 2 -> COUNT_DISABLED (does not return count at all)
+    history_count_mode: 1
```
jamesagnew, line 138: Would it be possible to just use the actual enum constants as the value here? eg. `COUNT_ACCURATE` instead of using number constants?

Dockerfile
```diff
@@ -36,7 +36,7 @@ COPY --from=build-hapi --chown=1001:1001 /tmp/hapi-fhir-jpaserver-starter/opente
 ENV ALLOW_EMPTY_PASSWORD=yes
 
 ########### distroless brings focus on security and runs on plain spring boot - this is the default image
-FROM gcr.io/distroless/java17-debian11:nonroot as default
+FROM gcr.io/distroless/java17-debian11:nonroot-arm64 as default
```
jamesagnew, line 39: Is this intentional? I think arm64 is stil the exception and not the rule...

sabukar-fn: Sorry this was meant to be a local MR, I'll remove this change for the real MR.

sabukar-fn: I'll close this and open another

sabukar-fn: Opened here https://github.com/hapifhir/hapi-fhir-jpaserver-starter/pull/500

Issue #433, opened 2022-09-28
Enable History Counting in application.yaml


ttran428:
Hi! I would like to request a feature to be able to configure `HistoryCountMode` in the `application.yaml` file: This is documented in https://hapifhir.io/hapi-fhir/docs/server_jpa/performance.html#history-counting which was created in https://github.com/hapifhir/hapi-fhir/issues/2641. Let me know if there is a different way I should be configuring this field!

PR #500, opened 2023-02-16
History count fix

sabukar-fn:
Resolves this issue https://github.com/hapifhir/hapi-fhir-jpaserver-starter/issues/433

During integration testing, it is useful to make rapid search history queries, but unless this flag can be configured we are forced to use the default of waiting 1 minute for the history cache to clear. Instead, during integration test we should always be getting the actual data from the db.

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -131,6 +131,11 @@ hapi:
       # These are allowed_origin patterns, see: https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/cors/CorsConfiguration.html#setAllowedOriginPatterns-java.util.List-
       allowed_origin:
         - '*'
+    # history_count_mode is an integer that represents the mode that the DAOConfig will use when getting the history count during a history request
+      # history_count_mode = 0 -> CACHED_ONLY_WITHOUT_OFFSET (default within hapi-fhir DAOConfig)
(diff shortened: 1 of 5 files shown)


src/main/resources/application.yaml
```diff
@@ -131,6 +131,11 @@ hapi:
       # These are allowed_origin patterns, see: https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/cors/CorsConfiguration.html#setAllowedOriginPatterns-java.util.List-
       allowed_origin:
         - '*'
+    # history_count_mode is an integer that represents the mode that the DAOConfig will use when getting the history count during a history request
+      # history_count_mode = 0 -> CACHED_ONLY_WITHOUT_OFFSET (default within hapi-fhir DAOConfig)
+      # history_count_mode = 1 -> COUNT_ACCURATE (ignores cache and always gets correct count)
+      # history_count_mode = 2 -> COUNT_DISABLED (does not return count at all)
+    history_count_mode: 1
```
jamesagnew, line 138: Instead of using numbers, why not just use the actual names? E.g.

```yaml
 history_count_mode: COUNT_ACCURATE
```

That way you don't have to have any map between the integer and the enum constant

jkiddo:
@sabukar-fn I'll close this due to inactivity. Feel free to reopen if you want to finalize the contribution at a later point in time.

PR #504, opened 2023-02-27
Update pom.xml

jkiddo:
Update to 6.4.1


tadgh:
Recommend you wait for the bump to 6.4.2 TBH. Should be out today.... 6.4.1 is an interim release to solve a really funky chicken and egg problem between core deps and HAPI-FHIR. That said, you can merge this if you like

tadgh (approved):

PR #520, opened 2023-04-13
Use JpaStorageSettings and StorageSettings

dotasek:
Refactor to use the renamed DaoConfig and ModelConfig classes.

This includes latest commits to master, and passes smoke tests.


michaelabuckley (approved):
Nice to keep up.  Doing this all at the end is a pain.

PR #558, opened 2023-06-26
Boyscouting

winne42:
Just some general boyscouting in preparation for a feature PR.

* Boyscouting: use List.of instead of Guava; use StandardCharsets instead of Guava; remove unused imports
* Boyscouting: typo in RepositoryValidationInterceptorFactoryDstu3
* Boyscouting: JUnit 5 conventions ("It is generally recommended to omit the public modifier for test classes, test methods, and lifecycle methods unless there is a technical reason for doing so)"
* Boyscouting: Docker conventions ("The instruction is not case-sensitive. However, convention is for them to be UPPERCASE to distinguish them from arguments more easily.")

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM docker.io/library/maven:3.9.2-eclipse-temurin-17 as build-hapi
+FROM docker.io/library/maven:3.9.2-eclipse-temurin-17 AS build-hapi
 WORKDIR /tmp/hapi-fhir-jpaserver-starter
 
 ARG OPENTELEMETRY_JAVA_AGENT_VERSION=1.26.0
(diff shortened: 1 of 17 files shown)


KevinDougan-SmileCDR:
Hi @winne42 ,

What do you mean by the term "Boyscouting" ?

Also, what problems are going to be solved by making these changes? There is no Issue associated with this PR so it is not clear what is being accomplished here.

Thanks!
Kevin

winne42:
Hello @KevinDougan-SmileCDR , thanks for the quick reaction!

What I meant was the Clean Code "[boy scout rule](https://matheus.ro/2017/12/11/clean-code-boy-scout-rule/)" where you want to leave code a just a little bit better/more tidy than you found it. As such, the PR does *not* want to solve any real problems. As you can see, I just improved slightly on following coding conventions (Docker, Java, JUnit 5), updated the code to use more modern Java features (`List.of()`, diamond operator, `StandardCharsets`), corrected some indentations etc. The PR should not have *any* functional changes to the code.

I am aware that different open source projects handle boy scout commits differently: some require a ticket for every PR, no matter how small it ist; some only accept PRs with functional changes; some don't want anyone to tidy up anything (yes, I experienced that too). I personally like to have boy scout commits in a separate PR because they typically touch many classes which can obscure more meaningful changes ("not seeing the forest for the trees").

If it better suits your workflow, I can of course open a ticket "just minor tidying up" and reference it in this PR.

Best regards, Winfried

jkiddo:
@winne42 I'm a total fan of this PR and the boyscout rule!

jkiddo (approved):

jkiddo:
@KevinDougan-SmileCDR - I'm not aware of any explicit rules for PR's having to reference an issue on this repo. That said, it might be a good idea. I'm fine merging this into main as it is simple, clean and makes the code base a better place ;)

PR #560, opened 2023-06-28
6.6.1 cr updates

Capt-Mac:
Added provider factory loaders for clinical reasoning operations as well as a delayed load as spring was not loading these beans in correct sequence to allow access to them.


jkiddo:
It looks like a lot of formatting went into this commit @Capt-Mac. Can you seperate the formatting from the 'business change in this'?

PR #568, opened 2023-07-21
add relaxedquery

delcroip:


jkiddo:
@delcroip its not clear what the intention is. Could you please elaborate?

jkiddo:
@delcroip closing due to inactivity

delcroip:
It was to allow $evaluate with get but despite the clear utility of it it was never approved so this changes can be closed

PR #577, opened 2023-08-17
Feature/using package installer spec

jkiddo:


XcrigX:
Are you missing this one in the change? 
https://github.com/hapifhir/hapi-fhir-jpaserver-starter/blob/master/src/main/java/ca/uhn/fhir/jpa/starter/common/StarterJpaConfig.java#L201

That was added so if you restart the server with the "implementationguides" section uncommented it won't re-install everything and start another re-index operation if search params are re-loaded.

jkiddo:
I don't think so. That would know be configurable - pr. IG as all fields on https://github.com/hapifhir/hapi-fhir/blob/a10856e0912df0758b202676d082ff4d01bef12f/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallationSpec.java#L35 can now be configured directly from the yaml

XcrigX:
I see. The defaults are different I think though. 
Starter defaults to NOT reload existing IGs, Core defaults to reload. 

Starter: https://github.com/hapifhir/hapi-fhir-jpaserver-starter/blob/master/src/main/java/ca/uhn/fhir/jpa/starter/AppProperties.java#L76

Core:  https://github.com/hapifhir/hapi-fhir/blob/a10856e0912df0758b202676d082ff4d01bef12f/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallationSpec.java#L86

Perhaps if you explicitly add "reloadExisting : false" into the examples in the yaml? 
At least commented out, but I'd advocate for defaulting it to false.

XcrigX:
Thanks!  That one bit me one time when a sever with lots of data restarted and I had a surprise reindex launch.

src/main/resources/application.yaml
```diff
@@ -76,14 +76,15 @@ hapi:
...
-    #    implementationguides:
+    implementationguides:
     ###    example from registry (packages.fhir.org)
-    #      swiss:
-    #        name: swiss.mednet.fhir
-    #        version: 0.8.0
+      swiss:
+        name: swiss.mednet.fhir
+        version: 0.8.0
+        reloadExisting : false
```
dotasek, line 84: @jkiddo Why is swiss.mednet.fhir changing from being left out to being included in the application?

jkiddo: No apparent reason - it was for test and I forgot to remove it again.

jkiddo:
I need someone to push the approve button before it can be merged

src/main/resources/application.yaml
patrick-werner: was commented out

PR #586, opened 2023-09-20
Test harness around use of binary storage interceptor

jkiddo:


jkiddo:
was part of https://github.com/hapifhir/hapi-fhir/issues/5333

PR #601, opened 2023-10-28
Feat/ig install

jkiddo:


jkiddo:
This is a feature requested by @costateixeira for speeding up the process of having an IG and make it usable without having to reboot the server instance. It is disabled by default.

dotasek:
Is there a way to automatically test this feature?

costateixeira:
Should be possible.
 upload a package.tgz - can be from any of our IGs (e.g. sample IG), do the
POST and then do the GET a conformance resource e.g.
StructureDefinition/myObservation

Not sure if we have to account for cache refresh etc.

dotasek (approved):
See testing procedure in comments.

costateixeira:
Sounds like "pass", right?

PR #605, opened 2023-11-12
Feature/dynamic content

jkiddo:
Refined how custom web apps can be hosted and how the standard templates can be changed using simple overwrites (FYI @costateixeira )


jkiddo:
@costateixeira you probably need to talk to some Smile guys if you want this into the starter project.

joofio:
Hi,
such a feature would be super welcomed to several projects!

costateixeira:
@dotasek probably you are expecting some testing behind this. I have tested and this works, but you may require some test case - is that true?
@jkiddo is this currently going to set the Flintstones image as the default logo? I would hope not.

jkiddo:
@costateixeira lets cook something different up in the next couple of days then

jkiddo:
@costateixeira I've fixed the defaults now to something more proper. Please let me know what else is on your mind.

jkiddo:
@dotasek & @costateixeira would you guys exchange thoughts on where this needs to head - down the drain or into master?

costateixeira:
I'd say master.
We already had some discussion on how to test this, right?

joofio:
i can help with testing.

jkiddo:
Sounds good to me - I would just like to not keep maintaining a separate branch but rather get it merged.

patrick-werner (approved):
looking great. I think the PR needs a better description as not only dynamic web_content support was added, but also operations for IG (un)installation.

jkiddo:
> looking great. I think the PR needs a better description as not only dynamic web_content support was added, but also operations for IG (un)installation.

IIRC then that has already be introduced. I'd say that this PR maybe did some unintended formatting around it.

PR #618, opened 2023-12-03
Update to HAPI FHIR 7.0.0

jamesagnew:
This PR is tracking the HAPI FHIR 7.0.x changes, using Spring 6 / Boot 3, and the new Jakarta APIs


jamesagnew:
Note: WebSockets don't currently work in this PR. Spring-Test 6.x requires Jetty 12, but Spring Boot 3.x seems to be incompatible with Jetty 12, so this is causing some weirdness as we've upgraded hapi-fhir to 12.

It looks like this is solved in Spring 6.1 so I'm going to try and get us up to that version soon.

jamesagnew:
Websockets are now enabled, and this branch has been upgraded to use Jetty 12

stmsat:
Hi James, I opened a merge request with your branch as the destination (#624) because I encountered issues when trying it (running spring-boot, tests, packaging a runnable war with at least tomcat, because jetty still has some issues), and had a somewhat hard time solving (or trying to solve) them; I hope you'll find it useful, in whole or part.

XcrigX:
I'm hitting an issue adding spring security to this branch. 
I haven't fully debugged it, but I _think_ the problem is that hapi-fhir-storage-cr has an explicit/older spring-security-core version: 

https://github.com/hapifhir/hapi-fhir/blob/master/hapi-fhir-storage-cr/pom.xml#L22
<spring-security-core.version>5.7.8</spring-security-core.version>

I'm seeing app context creation errors when adding my own _spring-boot-starter-security_ dependency to the starter project. I would guess SMILE would have similar issues? 

My two cents on the side, it might be much easier to maintain things if the HAPI root POM imported the SpringBoot BOM so all dependencies are vetted and compatible with a given SpringBoot version. HAPI is a super-complex project, so possible there are roadblocks to doing that, but.. maybe?

jkiddo:
@dotasek will this be merged anytime soon?

dotasek (approved):

PR #627, opened 2024-01-03
Explicit profile for smoke test invocation

dotasek:
The smoke tests appear to be failing due to an inability to start the server.

This explicitly sets the profile before running.


jkiddo (approved):
LGTM

Issue #631, opened 2024-01-19
Externalization of the application.yaml in Docker


psavva:
I would like to allow the externalization, and thus changes in configuration of the application.yaml configuration file.
This can be achieved using a bind volume.

@jkiddo

jkiddo:
Its supported already. There's an example on https://github.com/hapifhir/hapi-fhir-jpaserver-starter?tab=readme-ov-file#configuration-via-overridden-applicationyaml-file-and-using-docker

psavva:
Fair enough, closing

PR #632, opened 2024-01-19
Allow externalization of the application.yaml configuration file

psavva:
#631 Allow externatlization of application.yaml configuration file


psavva:
Build target also set upto Tomcat image as to avoid distroless for debugging and development testing....

PR #648, opened 2024-02-27
Update application.yaml

joofio:
to help with server.servlet.context-path:

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -1,3 +1,6 @@
+### context-path for using inside the HAPI Server. Server address will not apply any.
+#server.servlet.context-path: <path>
+
 #Adds the option to go to eg. http://localhost:8080/actuator/health for seeing the running configuration
 #see https://docs.spring.io/spring-boot/docs/current/reference/html/actuator.html#actuator.endpoints
 management:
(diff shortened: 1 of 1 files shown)


XcrigX:
I add this to all my deployments, so it makes sense to me to include it in the base application.yaml to clue other developers in on how to change it, but it's probably better to use the YAML style formatting to remain consistent with the rest of the properties. You could also leave it safely uncommented with a value such as: 
IE:
```
server: 
  servlet: 
    context-path : /

```

It might be nice to have a comment that will clue in someone on what the full path will actually be  - i.e. it will be a combination of server.servlet.context-path + /fhir. 
Maybe a comment along the lines of (not sure how best to word it) : 

```
# The base spring context path, things like the SpringBoot actuator will root off of this path. 
# The actual FHIR base url will use this path plus the value configured for the FHIR servletRegistrationBean in Application.java, which defaults to /fhir.
```

Side note, is it no longer possible to configure the base fhir path? Maybe it never was?
Appears to be hardcoded here:  
https://github.com/hapifhir/hapi-fhir-jpaserver-starter/blob/master/src/main/java/ca/uhn/fhir/jpa/starter/Application.java#L66

joofio:
your suggestion seems great!

jkiddo:
Documented on https://github.com/hapifhir/hapi-fhir-jpaserver-starter/blob/1e7690fc033326f7a9bd16c8f81dc44f8ab306f1/src/main/resources/application.yaml#L1 now - and it works as described and intended.

jkiddo:
was done in https://github.com/hapifhir/hapi-fhir-jpaserver-starter/pull/650

jkiddo:
@joofio / @XcrigX agree?

PR #654, opened 2024-02-29
Allow custom operations/providers in addition to interceptors

dconlan:
This change allows for a custom operation/provider to be added to the fhir server in the same way a custom interceptor is added.

### DIFF
--- a/README.md
+++ b/README.md
@@ -345,6 +345,16 @@ or
 
 2) classes will be instantiated via reflection if no matching Bean is found
 
+## Adding custom operations(providers)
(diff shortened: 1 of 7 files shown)


XcrigX:
looks good to me

XcrigX:
@dconlan - might be good to add a simple test as well. You can review CustomInterceptorTest to see how those were set up and place your custom provider into the existing some.custom.pkg1 package to make sure it's picked up and registered.

jkiddo (approved):

PR #657, opened 2024-03-06
Create extra-list.yaml

muhammad-levi:
Add ability to deploy extra manifests via the Helm chart (managed in the same file of values)


jkiddo:
@chgl can you have a look?

charts/hapi-fhir-jpaserver/templates/extra-list.yaml
chgl, line 1: This common library chart comes from bitnami, but isn't referenced in this chart so it won't work.

chgl, line 1: extraDeploy should be included in the values.yaml and the readme re-generated to update it

chgl (changes requested):

jkiddo:
Closed as deemed stale

PR #665, opened 2024-03-21
Harmonize use of InterceptorService

jkiddo:
Currently different two InterceptorService instances are used simultaneously - one from JpaConfig and one from RestfulServer. This might cause confusion and does not seems necessary.


tadgh:
Please have a look at CompositeInterceptorBroadcaster to have a look at how pointcut invocations are spread across multiple interceptor services. Yes, there are intentionally 2 in HAPI-FHIR, because we occasionally have a FHIR Endpoint with no persistence layer, and occasionally have a persistence layer with no FHIR Endpoint. The two layers have separate pointcuts, and the compositeinterceptorbroadcaster is what allows you to span the chasm and invoke pointcuts on the other layer if needed. FWIW, since in hapi-fhir-jpaserver-starter this is all one big layer, i think it's fine to have a single interceptor service.

tadgh (approved):

tadgh (changes requested):
.

jkiddo:
Thx for the clarifying response. Wouldn't a check on instance equality on broadcasters on e.g.  https://github.com/hapifhir/hapi-fhir/blob/4a726dd2d1d015ad03340d9c726a31c2b0ee831b/hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/util/CompositeInterceptorBroadcaster.java#L83 fix that so it isn't invoked twice?

jkiddo:
( including `AopUtils.isAopProxy` ) that is or something like that ...

jkiddo:
I get the design with the flexibility that sometimes it is used as a facade and sometime its more mainly a storage setup - but in the pure starter setup, wouldn't the above suggestion fix that and also make it more simple in total? With the current setup you need to be extra cautious about _where_ you register interceptors.

jkiddo:
@XcrigX try change https://github.com/hapifhir/hapi-fhir-jpaserver-starter/blob/624a269994933cc51a7b269f33e9f2006c7954a7/src/test/java/some/custom/pkg1/CustomInterceptorPojo.java#L14 to `@Hook(Pointcut.STORAGE_PRESTORAGE_RESOURCE_CREATED)` and you'll see the difference

jkiddo:
I got it wrong. It works just as @tadgh described it, though the double invocation could be solved if instance equality is checked.

PR #660, opened 2024-03-11
HAPI 7.2.0 Release Tracking

dotasek:
This branch brings JPA server starter up to date with HAPI release 7.2.0

PR #672, opened 2024-04-01
671 - Updated the run instructions.

### LINKED ISSUE #671: Update Run Instructions
This is based on reported issue #667 .

This issue is to update the instructions on how to run the server, as the [current instructions in the pom.xml file](https://github.com/hapifhir/hapi-fhir-jpaserver-starter/blob/master/pom.xml#L600) are not correct.

KevinDougan-SmileCDR:
Closes #671

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,6 +191,13 @@ public class YourInterceptor
 
 The easiest way to run this server entirely depends on your environment requirements. The following ways are supported:
 
+### Using jetty
+```bash
+mvn -Pjetty spring-boot:run
+```
+
+The Server will then be accessible at http://localhost:8080/fhir and the CapabilityStatement will be found at http://localhost:8080/fhir/metadata.
+
 ### Using Spring Boot
 ```bash
 mvn spring-boot:run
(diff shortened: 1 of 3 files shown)


dotasek (changes requested):
I came across this request because my standard jetty start method was failing in the HAPI release tracking PR: https://github.com/hapifhir/hapi-fhir-jpaserver-starter/pull/660

I can confirm that all three of these methods get me a working instance which passes all the smoke tests.

However, I believe this means we should also replace the content in the README, some of which no longer works: https://github.com/hapifhir/hapi-fhir-jpaserver-starter?tab=readme-ov-file#using-jetty

Could you update the README as well?

jkiddo:
https://github.com/hapifhir/hapi-fhir-jpaserver-starter/blob/106af029a25f8ba082aabf3c599d9128c5e90db4/README.md?plain=1#L203 needs updating @KevinDougan-SmileCDR

KevinDougan-SmileCDR:
Thanks @jkiddo and @dotasek , I will also update the README, but I've been pulled into a High Priority issue for the time being. I'll complete this soon...

README.md
```diff
@@ -189,21 +189,27 @@ public class YourInterceptor
...
+### Using jetty
+```bash
+mvn -Pjetty spring-boot:run
+```
 
 If you need to run this server on a different port (using Maven), you can change the port in the run command as follows:
 
 ```bash
-mvn -Pjetty -Djetty.port=8888 jetty:run
+mvn -Pjetty -Dserver.port=8888 spring-boot:run
```
dotasek, line 209: This doesn't appear to work as advertised. If I run in using these params, I still get the server on port 8080.

dotasek (changes requested):
Port selection instructions do not appear to work as written.

README.md
KevinDougan-SmileCDR: @dotasek   Yes, that is why I haven't finished work on this PR yet. That parameter is supposed to work, but it doesn't have any effect. I will keep investigating this to try and figure it out.

XcrigX: Not sure if helpful, but in my deployments of HAPI I set the port in the application.yml like this and it works: 

```
server: 
  servlet: 
    context-path : /
  port: 8443
```

If that does work, I might suggest just adding that section the default property file (using 8080 instead of 8443 obviously).

XcrigX: I misread that the problem is for Jetty, which I'm not using. 
But I notice it might be server.servlet.port and not just server.port.

KevinDougan-SmileCDR: @XcrigX   I have updated the instructions and added an example as per your recommendation...thanks for that!

KevinDougan-SmileCDR:
@dotasek   Please approve if you are satisfied with the latest changes.

dotasek (approved):
New instructions for port work with smoke tests. Looks good. Thanks!

PR #676, opened 2024-04-13
Add profile for GCP CloudSql Postgres driver

jakubadamek:


jkiddo:
Please add a comment in the poem that this profile also needs eg. the `boot` profile to work

jakubadamek:
> Please add a comment in the poem that this profile also needs eg. the `boot` profile to work

Thanks, done

dotasek:
Is there a way to test what this new dependency adds to the functionality? (example configs, start params, etc.?)

pom.xml
```diff
@@ -676,5 +676,16 @@
                 </plugins>
             </build>
         </profile>
+        <!-- For connecting to GCP CloudSQL Postgres instances. Needs 'boot' profile as well. -->
```
jkiddo, line 679: You might consider putting the link you found here as well (the https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/main/docs/jdbc.md#postgres-1 one)

jakubadamek: Done

jkiddo (approved):

Issue #669, opened 2024-03-30
Property java.version=11


netlogo-it:
Spring Boot 3.2.0 requires Java 17 as a minimum version. In the pom.xml file version 11 is specified, I think this is an oversight, probably the java.version property is not used. I think the property should be removed.

madduci:
@jkiddo here i've made a PR for it: https://github.com/hapifhir/hapi-fhir-jpaserver-starter/pull/686

PR #686, opened 2024-05-24
Updated Java Minimum version to 17

madduci:
Resolves issue #669

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
-FROM docker.io/library/maven:3.9.4-eclipse-temurin-17 AS build-hapi
+FROM docker.io/library/maven:3.9.6-eclipse-temurin-17 AS build-hapi
 WORKDIR /tmp/hapi-fhir-jpaserver-starter
 
-ARG OPENTELEMETRY_JAVA_AGENT_VERSION=1.31.0
+ARG OPENTELEMETRY_JAVA_AGENT_VERSION=1.33.3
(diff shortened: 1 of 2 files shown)


jkiddo (changes requested):
Overall - looks fine and safe since HAPI FHIR uses Spring Boot v3.2 declared here: https://github.com/hapifhir/hapi-fhir/blob/0fe3380c4a613fcb5c4af99a228bfcc7971f162a/pom.xml#L1013 

It does also requires a bump on tomcat here: https://github.com/hapifhir/hapi-fhir-jpaserver-starter/blob/8f497dd7ab8d8372ec1bc9e8144d363cfc40f7eb/Dockerfile#L21 due to https://spring.io/blog/2022/11/16/spring-framework-6-0-goes-ga

jkiddo:
@jamesagnew I imagine you would have an opinion on the timing of this?

madduci:
I have a question about it. Should we swap the Tomcat image against e.g. Eclipse-Temurin base image? It comes also with a shell for debugging purposes and can run the SpringBoot package like the distroless one (also the Alpine one)

From one side it's more resilient compared to the one of Tomcat, maybe also smaller. On the other side, Tomcat should be tested separately, but for Production use in environments such as Kubernetes, I see that smaller attack surfaces are better.

jkiddo:
Main reason for having tomcat was to be sure that the application was compatible with tomcat. I dont know who uses the dockerized version of the application inside tomcat inside Docker.

PR #739, opened 2024-09-19
History count fix

sabukar-fn:
Reopening this old PR after addressing changes https://github.com/hapifhir/hapi-fhir-jpaserver-starter/pull/500

Resolves this issue https://github.com/hapifhir/hapi-fhir-jpaserver-starter/issues/433

During integration testing, it is useful to make rapid search history queries, but unless this flag can be configured we are forced to use the default of waiting 1 minute for the history cache to clear. Instead, during integration test we should always be getting the actual data from the db.


src/main/java/ca/uhn/fhir/jpa/starter/common/StarterJpaConfig.java
```diff
@@ -63,6 +64,8 @@
 import ca.uhn.fhir.validation.IValidatorModule;
 import ca.uhn.fhir.validation.ResultSeverityEnum;
 import com.google.common.base.Strings;
+import com.google.common.collect.ImmutableList;
```
sabukar-fn, line 67:
```suggestion
```

src/main/java/ca/uhn/fhir/jpa/starter/AppProperties.java
```diff
@@ -71,6 +71,7 @@ public class AppProperties {
   private List<String> supported_resource_types = new ArrayList<>();
   private List<Bundle.BundleType> allowed_bundle_types = null;
   private Boolean narrative_enabled = true;
+  private String history_count_mode = "CACHED_ONLY_WITHOUT_OFFSET";
```
jkiddo, line 74: This should be bound to HistoryCountModeEnum like in https://github.com/hapifhir/hapi-fhir-jpaserver-starter/blob/e2d2ea7726975e76010fc3f98149fad72a48fe84/src/main/java/ca/uhn/fhir/jpa/starter/AppProperties.java#L67 making it more strongly typed

jkiddo (changes requested):
The binding to the enum should be strongly typed which would influence this entire PR

PR #742, opened 2024-09-25
Expose Setting for prefetch thresholds

tadgh:
Related to discussion here: https://github.com/hapifhir/hapi-fhir/issues/6109 

- expose prefetch threshold setting
- add docs

@bashir2

### DIFF
--- a/src/main/java/ca/uhn/fhir/jpa/starter/AppProperties.java
+++ b/src/main/java/ca/uhn/fhir/jpa/starter/AppProperties.java
@@ -104,7 +104,7 @@ public class AppProperties {
 
 	private final List<String> custom_provider_classes = new ArrayList<>();
 
-	private List<Integer> search_prefetch_thresholds = List.of(13, 503, 2003, -1);
+	private List<Integer> search_prefetch_thresholds = new ArrayList<>();
 
 
 	public List<String> getCustomInterceptorClasses() {
--- a/src/main/java/ca/uhn/fhir/jpa/starter/AppProperties.java
+++ b/src/main/java/ca/uhn/fhir/jpa/starter/AppProperties.java
@@ -104,7 +104,7 @@ public class AppProperties {
 
 	private final List<String> custom_provider_classes = new ArrayList<>();
 
-	private List<Integer> search_prefetch_thresholds = new ArrayList<>();
+	private List<Integer> search_prefetch_thresholds = List.of(13, 503, 2003, -1);
 
 
 	public List<String> getCustomInterceptorClasses() {
--- a/src/main/java/ca/uhn/fhir/jpa/starter/AppProperties.java
+++ b/src/main/java/ca/uhn/fhir/jpa/starter/AppProperties.java
@@ -104,6 +104,8 @@ public class AppProperties {
 
 	private final List<String> custom_provider_classes = new ArrayList<>();
 
+	private List<Integer> search_prefetch_thresholds = new ArrayList<>();
+
 
 	public List<String> getCustomInterceptorClasses() {
     return custom_interceptor_classes;
@@ -643,7 +645,16 @@ public Cors getCors() {
 		this.language_search_parameter_enabled = language_search_parameter_enabled;
 	}
 
-	public static class Cors {
+	public List<Integer> getSearch_prefetch_thresholds() {
+		return this.search_prefetch_thresholds;
+	}
+
+	public void  setSearch_prefetch_thresholds(List<Integer> thePrefetchThresholds) {
+		this.search_prefetch_thresholds = thePrefetchThresholds;
+	}
+
+
+    public static class Cors {
     private Boolean allow_Credentials = true;
     private List<String> allowed_origin = List.of("*");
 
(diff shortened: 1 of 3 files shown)


src/main/resources/application.yaml
```diff
@@ -76,7 +76,7 @@ hapi:
     ### This flag when enabled to true, will avail evaluate measure operations from CR Module.
     ### Flag is false by default, can be passed as command line argument to override.
     cr:
-      enabled: false
+      enabled: true
```
codeforgreen, line 79: **question**: is this an intentional change? I didn't see how it relates to the reported issue.

codeforgreen (approved):

src/main/resources/application.yaml
tadgh: ah good catch, not intentional

src/main/resources/application.yaml
```diff
@@ -229,7 +229,18 @@ hapi:
...
+    # Search Prefetch Thresholds.
+
+    # This setting sets the number of search results to prefetch. For example, if this list
+    # is set to [100, 1000, -1] then the server will initially load 100 results and not
+    # attempt to load more. If the user requests subsequent page(s) of results and goes
+    # past 100 results, the system will load the next 900 (up to the following threshold of 1000).
+    # The system will progressively work through these thresholds.
+    # A threshold of -1 means to load all results. Note that if the final threshold is a
+    # number other than -1, the system will never prefetch more than the given number.
+    search_prefetch_thresholds: 13,503,2003
```
bashir2, line 242: Based on the above comment, shouldn't this list end with `-1` (similar to the [default](https://github.com/hapifhir/hapi-fhir/blob/20d3e6bb252eff52def255669ca6f590c2d8db59/hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/api/config/JpaStorageSettings.java#L191))?

tadgh: Ah thank you, yes, you are correct. patched

tadgh:
@bashir2 this look better?

src/main/java/ca/uhn/fhir/jpa/starter/AppProperties.java
```diff
@@ -104,6 +104,8 @@ public class AppProperties {
 
 	private final List<String> custom_provider_classes = new ArrayList<>();
 
+	private List<Integer> search_prefetch_thresholds = List.of(13, 503, 2003, -1);
```
bashir2, line 107: nit: Another option is not to call `jpaStorageSettings.setSearchPreFetchThresholds` if the list is not provided. I am just thinking about the edge-case where this goes out of sync from the the actual default [here](https://github.com/hapifhir/hapi-fhir/blob/20d3e6bb252eff52def255669ca6f590c2d8db59/hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/api/config/JpaStorageSettings.java#L191).

bashir2 (approved):
Thanks @tadgh for the quick fix.

src/main/java/ca/uhn/fhir/jpa/starter/AppProperties.java
tadgh: Good call! I have patched it.

src/main/resources/application.yaml
```diff
@@ -229,7 +229,18 @@ hapi:
...
+    # Search Prefetch Thresholds.
+
+    # This setting sets the number of search results to prefetch. For example, if this list
+    # is set to [100, 1000, -1] then the server will initially load 100 results and not
+    # attempt to load more. If the user requests subsequent page(s) of results and goes
+    # past 100 results, the system will load the next 900 (up to the following threshold of 1000).
+    # The system will progressively work through these thresholds.
+    # A threshold of -1 means to load all results. Note that if the final threshold is a
+    # number other than -1, the system will never prefetch more than the given number.
+    search_prefetch_thresholds: 13,503,2003,-1
```
jkiddo, line 242: If these are the default values shouldn't they be provided as such in the code here https://github.com/hapifhir/hapi-fhir-jpaserver-starter/pull/742/files#diff-79251bbeac3bb7f417a707bbda464fdfb2394f40bd88e43a9f5472e0ed351c83R107?

tadgh: Nah its fine, this is only populated if non-empty, and will rely on the HAPI-FHIR defaults if absent/empty. I put these in the properties file here for discoverability only. I hesitate to make them a java default in jpaserver-starter for fear of drift between hapi and this project.

dotasek (approved):
Looks good to me. Tracing startup shows the settings being set via jpaStorageSettings.setSearchPreFetchThresholds in FhirServerConfigCommon

Can confirm what @tadgh was saying with regard to defaults; if the setting is empty in application.yaml, the default values from HAPI FHIR are used (link below)

https://github.com/hapifhir/hapi-fhir/blob/71cb097aa647657e7073990dc16d90d566020bb4/hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/api/config/JpaStorageSettings.java#L192

jkiddo (approved):

PR #745, opened 2024-09-29
Feature/fixing prometheus

jkiddo:
[Fixing prometheus so that it actually works. Also seems related to](https://github.com/hapifhir/hapi-fhir-jpaserver-starter/commit/94aa71f894c012320d5f1dfef2a4036bfdd307b7) https://github.com/micrometer-metrics/micrometer/issues/5093

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -11,7 +11,7 @@ management:
     enabled-by-default: false
     web:
       exposure:
-        include: 'health,info,prometheus,metrics' # or '*' for all'
+        include: 'health,info' # or e.g. 'health,info,prometheus,metrics' or '*' for all'
   endpoint:
     metrics:
       enabled: true
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -1,31 +1,35 @@
 #Uncomment the "servlet" and "context-path" lines below to make the fhir endpoint available at /example/path/fhir instead of the default value of /fhir
 server:
-#  servlet:
-#    context-path: /example/path
+  #  servlet:
+  #    context-path: /example/path
   port: 8080
 #Adds the option to go to eg. http://localhost:8080/actuator/health for seeing the running configuration
 #see https://docs.spring.io/spring-boot/docs/current/reference/html/actuator.html#actuator.endpoints
 management:
+  endpoints:
+    enabled-by-default: false
+    web:
+      exposure:
+        #include: 'health,info,prometheus' # or '* for all'
+        include: '*'
   endpoint:
-    endpoints:
-      enabled-by-default: false
-      web:
-        exposure:
-          include: health,prometheus
+    metrics:
+      enabled: true
     health:
       enabled: true
       probes:
         enabled: true
-      livenessState:
-        enabled: true
-      readinessState:
-        enabled: true
+      group:
+        liveness:
+          include:
+            - livenessState
+            - readinessState
     prometheus:
       enabled: true
-      metrics:
-        export:
-          enabled: true
-
+  prometheus:
+    metrics:
+      export:
+        enabled: true
 spring:
   main:
     allow-circular-references: true
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -6,7 +6,7 @@ server:
 #Adds the option to go to eg. http://localhost:8080/actuator/health for seeing the running configuration
 #see https://docs.spring.io/spring-boot/docs/current/reference/html/actuator.html#actuator.endpoints
 management:
-  #The following configuration will enable the actuator endpoints at /actuator/health, /actuator/info, /actuator/prometheus
+  #The following configuration will enable the actuator endpoints at /actuator/health, /actuator/info, /actuator/prometheus, /actuator/metrics. For security purposes, only /actuator/health is enabled by default.
   endpoints:
     enabled-by-default: false
     web:
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -11,7 +11,7 @@ management:
     enabled-by-default: false
     web:
       exposure:
-        include: 'info,health' # or e.g. 'info,health,prometheus,metrics' or '*' for all'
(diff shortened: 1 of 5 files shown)


dotasek (changes requested):
What is the testable result of this change?

From the Prometheus docs, when this is running, there should be an endpoint available that exposes metrics at `/actuator/prometheus` or similar. If this is enabled by default, we could also include a smoke test to verify that this is running.

jkiddo:
@dotasek I've added this: https://github.com/hapifhir/hapi-fhir-jpaserver-starter/pull/745/commits/dfce153db1d877b4c86af809fb2705101deb9447

jkiddo:
... and now a smoketest for the promethus endpoint as well

dotasek:
This looks good so far. When I get a chance later today, I'll spin up the project and check if it's reporting metrics too, and approve if everything looks OK.

Thanks Jens!

dotasek (changes requested):
I do not see `actuator/info` appearing as part of the default config.

`prometheus`, `health`, and `metrics` appear, but I need to explicitly set up `info` to get that endpoint, in both the main and test application configs:

```yaml
  endpoint:
    ...
    info:
      enabled: true
    ...
```

Is leaving `info` out intended?

It would also not take much to extend that IT test to cover all these endpoints. I tried the following, and it worked for me:

```java
        @ParameterizedTest
	@ValueSource(strings = {"prometheus", "info", "health", "metrics"})
	void testActuatorEndpointExists(String endpoint) throws IOException {

		CloseableHttpClient httpclient = HttpClients.createDefault();
		CloseableHttpResponse response = httpclient.execute(new HttpGet("http://localhost:" + port + "/actuator/" + endpoint));
		int statusCode = response.getStatusLine().getStatusCode();
		assertEquals(200, statusCode);

	}
```

Could you add that as well?

XcrigX:
Does the prometheus endpoint expose anything potentially sensitive? 
If I read this correctly, only /health is enabled by default over HTTP: https://docs.spring.io/spring-boot/reference/actuator/endpoints.html#actuator.endpoints.security

dotasek:
@jkiddo I think @XcrigX may have a very valid point. This should be most secure by default, and we would have to review each of these endpoints to check what HAPI produces.

Maybe it's best to only configure /health, and comment the rest with a warning about security concerns.

XcrigX:
> @jkiddo I think @XcrigX may have a very valid point. This should be most secure by default, and we would have to review each of these endpoints to check what HAPI produces.
> 
> Maybe it's best to only configure /health, and comment the rest with a warning about security concerns.

Agree with this.  I'd include them commented out so it's easy to enable, but not as default.  The good news is that base for these is at /actuator which is a sibling of the /fhir baseurl.  So it's easy to enable them internally and prevent public access through a reverse proxy/gateway.

XcrigX:
I don't disagree with you in general, @jkiddo.  But consider this scenario.  I am deploying my own application.yml file as a file with my server deployment. My application.yml doesn't mention the actuator endpoints at all.

Tomorrow, I pull the latest code and build it which now includes these options in the classpath copy of application.yaml.  I re-deploy my server (not changing my locally deployed properties file), and now the endpoints are suddenly enabled because these properties are not explicitly overridden in my deployed properties but are present on the classpath.

XcrigX:
Here are the relevant docs on property loading order: 
https://docs.spring.io/spring-boot/reference/features/external-config.html

I'm pretty sure any property in the compiled/jar copy of the properties file will but picked up if it's not explicitly overridden in a deployment - so it makes sense to me to be careful and only have safe/reasonable defaults enabled.

jkiddo:
@XcrigX - that's a fair point, though I experience a different behavour. When using `--spring.profiles.active=diff` `--spring.config.location=...application-diff.yaml` whatever is stated in that config file is what gets loaded. It does not take into account what was stated in the bundled one unless I'm doing something wrong.

XcrigX:
I just ran a quick test to verify my understanding. Here's what I did. 
1) Pulled latest master branch. 
2) Updated the application.yaml, changed the server.port property to 8888. 
3) compiled w/ "mvn clean package spring-boot:repackage -DskipTests=true -Pboot"
4) took the compiled uber jar - ROOT.war file, copied it to a different directory on my machine
5) copied the application.yaml file to the same directory, edited it and commented out the server.port property. 
6) ran the server (java -jar ROOT.war) 

The server came up on 8888, despite my local application.yaml not specifying that port and it not being the Spring default- it's picking up the default value specified in the compiled jar. 

To double-check myself I then stopped the server, uncommented server.port and changed it to 8081, reran and it came up on 8081.

jkiddo:
I guess that the reason that you @XcrigX experience a different behaviour than me is because I use the `spring.config.location`:

image

dotasek:
@jkiddo but why re-introduce `/info`? I thought only `/health` was recommended?

jkiddo:
@dotasek that was a blunder from my side. It's fixed now.

src/main/resources/application.yaml
```diff
@@ -1,31 +1,37 @@
...
 server:
-#  servlet:
-#    context-path: /example/path
+  #  servlet:
+  #    context-path: /example/path
   port: 8080
 #Adds the option to go to eg. http://localhost:8080/actuator/health for seeing the running configuration
 #see https://docs.spring.io/spring-boot/docs/current/reference/html/actuator.html#actuator.endpoints
 management:
+  #The following configuration will enable the actuator endpoints at /actuator/health, /actuator/info, /actuator/prometheus
```
dotasek, line 9: One minor nitpick. I think this would make the default config more obvious.
 
  #The following configuration will enable the actuator endpoints at /actuator/health, /actuator/info, /actuator/prometheus, /actuator/metrics. For security purposes, only /actuator/health is enabled by default.

dotasek (changes requested):
One minor nitpicky change, and I think this is done. As always, thanks for the patience, Jens.

src/main/resources/application.yaml
jkiddo: done

dotasek (approved):

PR #748, opened 2024-10-04
Improve Stale management

joofio:
In order to make the most of our time managing issues, i am suggesting to reduce drastically the time for tagging as stale and increasing the time for comment / update before closing time.
I would also add a "whitelist" label so it is not affected by it ("bug")

### DIFF
--- a/.github/workflows/stale.yaml
+++ b/.github/workflows/stale.yaml
@@ -9,11 +9,12 @@ jobs:
     steps:
       - uses: actions/stale@v9
         with:
-          stale-issue-message: 'This issue is stale because it has been open 180 days with no activity. Remove stale label or comment or this will be closed in 5 days.'
+          stale-issue-message: 'This issue is stale because it has been open 90 days with no activity. Remove stale label or comment or this will be closed in 10 days.'
           stale-pr-message: 'This PR is stale because it has been open 45 days with no activity. Remove stale label or comment or this will be closed in 10 days.'
           close-issue-message: 'This issue was closed because it has been stalled for 5 days with no activity.'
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yaml
```diff
@@ -9,11 +9,12 @@ jobs:
     steps:
       - uses: actions/stale@v9
         with:
-          stale-issue-message: 'This issue is stale because it has been open 730 days with no activity. Remove stale label or comment or this will be closed in 5 days.'
+          stale-issue-message: 'This issue is stale because it has been open 90 days with no activity. Remove stale label or comment or this will be closed in 10 days.'
           stale-pr-message: 'This PR is stale because it has been open 45 days with no activity. Remove stale label or comment or this will be closed in 10 days.'
           close-issue-message: 'This issue was closed because it has been stalled for 5 days with no activity.'
```
XcrigX, line 14: I think this line needs to be updated also to reflect 10

XcrigX: ..line 14 "close-issue-message" line

joofio: done

jkiddo (approved):

jkiddo:
@dotasek also good for you?

dotasek (changes requested):
The general consensus from the HAPI side of things, via @michaelabuckley and @jamesagnew, is that a 3 month span is far too short, particularly since both projects have a quarterly release cycle.

I think a review of how long it takes issues to be resolved would give us a better idea of 'how short is too short'.

jkiddo:
The starter project releases are not necessarily bound to the parent project as features have been added over time independent of the parent - just saying.

It is however convenient that major/minor matches between the projects.

PR #776, opened 2025-01-28
Update fhir version

kennankole:
Updating FHIR version from `R4` to `R5`


XcrigX:
I agree. R4 is the defacto industry standard right now and what most people use. Updating the project default could have downstream consequences for folks/processes that aren't explicitly overriding the properties.

kennankole:
It was a mistake on my part, sorry 🤭

PR #785, opened 2025-02-24
Update to HAPI FHIR 8.2.0

dotasek:
Update to HAPI FHIR 8.2.0

PR #787, opened 2025-03-04
Revision Number Versioning

dotasek:
This adds a revision number to the maven version for this project.

All releases from this point forward will be defined as a tagged commit in the master branch with a corresponding push of the project Docker images to DockerHub.

Each version of jpaserver-starter will be based on the HAPI release version it is based on, with a revision number.

**Format:**

`{HAPI-VERSION}-{REVISION-NUMBER}`

**Example:**

`8.0.0-1`

The revision number will increment when a new jpaserver-starter is released using the same HAPI version as its base.  Thus a sequence of releases based starting with HAPI 8.0.0 would look as follows:

* 8.0.0-1 (first release using HAPI 8.0.0)
* 8.0.0-2 (bug fix in jpaserver-starter)
* 8.0.0-3 (new config options provided in jpaserver-starter)

* 8.2.0-1 (first release using HAPI 8.2.0)
...

Note that there is no distinction between features or bug-fixes in the revision version.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,6 +3,11 @@
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
+    <properties>
+        <java.version>17</java.version>
+        <hapi.fhir.jpa.server.starter.revision>1</hapi.fhir.jpa.server.starter.revision>
+    </properties>
+
     <!-- one-liner to take you to the cloud with settings form the application.yaml file: -->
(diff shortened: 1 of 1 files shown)


dotasek:
> I like this scheme, FWIW. Curious, will we still still bump up to SNAPSHOT versions of the parent between releases? i.e. would we have something like : 8.1.1-SNAPSHOT-1?

Yes, exactly that. I have an update to the current release tracking branch (https://github.com/hapifhir/hapi-fhir-jpaserver-starter/pull/785) that I'm waiting to commit and push.

We discussed this with @JPercival and @jamesagnew. The PR description is what we came up with. I'm just awaiting approval here before following up with updating the release tracker, writing documentation and releasing (8.0.0-1 for the first one, since we have 8.0.0 already).

JPercival (approved):

Issue #788, opened 2025-03-07
$diff not supported


pdavy-nhs:
**Description**

If I try to do a **/$diff** e.g. `fhir/Task/64e31f69-31cf-48cb-8bcc-e1669a79687f/$diff` I get

```{
    "resourceType": "OperationOutcome",
    "issue": [
        {
            "severity": "error",
            "code": "not-supported",
            "diagnostics": "Invalid request: The FHIR endpoint on this server does not know how to handle GET operation[Task/64e31f69-31cf-48cb-8bcc-e1669a79687f/$diff] with parameters [[]]"
        }
    ]
}```

The docs suggest $diff has been supported for some time

**Version:** JPA Starter 7.2.1

XcrigX:
Interesting. It works on the public test server.  Ex: https://hapi.fhir.org/baseR4/Task/641461/$diff

But I tried against a local instance of HAPI JPA and get the same response as you. I don't see an obvious way to "turn it on" in the properties.

pdavy-nhs:
> Interesting. It works on the public test server. Ex: https://hapi.fhir.org/baseR4/Task/641461/$diff
> 
> But I tried against a local instance of HAPI JPA and get the same response as you. I don't see an obvious way to "turn it on" in the properties.

Yes, I read some older issue where `DiffProvider` needed to be injected, I can't see any sign of that class in 7.2 or the latest version (and want to avoid forking the jpa server anyway). Also, my capability statement doesn't reference the diff operation whereas the public test server does.

RaviNatarajan22:
Was it accidently got pulled out from the release build ?

XcrigX:
If you can figure out which class needs to be registered, I'm sure a PR to add back it would be welcomed.

pdavy-nhs:
> If you can figure out which class needs to be registered, I'm sure a PR to add back it would be welcomed.

I believe it's this 
https://hapifhir.io/hapi-fhir/apidocs/hapi-fhir-jpaserver-base/ca/uhn/fhir/jpa/provider/DiffProvider.html
https://hapifhir.io/hapi-fhir/apidocs/hapi-fhir-jpaserver-base/src-html/ca/uhn/fhir/jpa/provider/DiffProvider.html

But the source doesn't seem to exist in the repo (apologies I have no experience with the inner workings of hapi fhir)

XcrigX:
[...]

PR #791, opened 2025-03-12
registered  diff provider

XcrigX:
Diff provider was not registered by default. Added it and a test. See #788


XcrigX:
image

Every time I submit a PR, I seem to get this failure.  I'm not sure what causes it, or even what it means. It looks like the spotless check passed the second time around, I don't know what this _mshick_ thing is trying to do.

XcrigX:
Thanks [jkiddo](https://github.com/jkiddo) - Do you know what's going on with the github actions check failing (see my message/screenshot above).  Am I doing something incorrectly when I submit PRs, or are the actions broken?

jkiddo:
@XcrigX nope - I see it myself as well. We probably should look into it.

PR #795, opened 2025-03-19
Added support for hybrid remote terminology

jkiddo:

PR #801, opened 2025-04-16
fix: set default value for request_tenant_partitioning_mode to true to support easy default partitioning

patrick-werner:
### Description

This PR updates the `hapi-fhir-jpaserver-starter` to set `request_tenant_partitioning_mode = true` by default in AppProperties & `application.yaml`.  
This change ensures that multi-tenancy is enabled using the tenant ID in the request URL base, which is handled by the following logic:

```java
if (myAppProperties.getPartitioning().getRequest_tenant_partitioning_mode() == Boolean.TRUE) {
	ourLog.info("Partitioning mode enabled in: Request tenant partitioning mode");
	myRestfulServer.registerInterceptor(new RequestTenantPartitionInterceptor());
	myRestfulServer.setTenantIdentificationStrategy(new UrlBaseTenantIdentificationStrategy());
}

### DIFF
--- a/src/main/java/ca/uhn/fhir/jpa/starter/AppProperties.java
+++ b/src/main/java/ca/uhn/fhir/jpa/starter/AppProperties.java
@@ -855,7 +855,7 @@ public class AppProperties {
 		private Boolean database_partition_mode_enabled = false;
 		private Boolean patient_id_partitioning_mode = false;
 		private Integer default_partition_id = 0;
-		private boolean request_tenant_partitioning_mode;
+		private boolean request_tenant_partitioning_mode = true;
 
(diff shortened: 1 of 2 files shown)


jkiddo (approved):
I would favour the toggle to be something along how its done here https://github.com/hapifhir/hapi-fhir-jpaserver-starter/pull/795/files#diff-49a09f45e926be807a51048b2998d1ce0341571cfa6bc23927286d17c06fd1a5R17 but that is just my opinion, as I find it a bit un-intuitive that you both have to add the configuration values **AND** have to flag them as enabled ... but maybe its something  we can take at DevDays or something (@jamesagnew)

PR #803, opened 2025-04-18
Fixed bug in setting of jpaStorageSettings pre-expansion settings

sliver007:
Because of the way the setters in JpaStorageSettings are defined for PreExpandValueSetsMaxCount and PreExpandValueSetsDefaultCount, if PreExpandValueSetsDefaultCount is set before PreExpandValueSetsMaxCount, PreExpandValueSetsDefaultCount will be reset to its default value by the setter for PreExpandValueSetsMaxCount.
This does not do anything to change the setters, but switches the order in which they are called from FhirServerConfigCommon.


robogary:
Formatting check succeeded!

PR #822, opened 2025-05-11
Moved terminology config out of being R4 specific. It now applies to …

jkiddo:
…all versions


robogary:
Formatting check succeeded!

patrick-werner (approved):

Issue #807, opened 2025-04-24
Enable full text search in docker container


Yes25:
I tried to enable full text search with the prebuilt docker image and Lucene but can't get it to work. 

I tried this compose file:
```yaml
services:
  fhir:
    container_name: fhir
    image: "hapiproject/hapi:latest"
    ports:
      - "8080:8080"
    configs:
      - source: hapi
        target: /app/config/application.yaml

configs:
  hapi:
     file: ./hapi.application.yaml
```

along with this application.yaml:
```yaml
spring:
  datasource:
    url: 'jdbc:h2:file:./target/database/h2'
    #url: jdbc:h2:mem:test_mem
    username: sa
    password: null
    driverClassName: org.h2.Driver
    max-active: 15

    # database connection pool size
    hikari:
      maximum-pool-size: 10
  jpa:
    properties:
      hibernate.dialect: ca.uhn.fhir.jpa.model.dialect.HapiFhirH2Dialect
      hibernate.search.enabled: true
      # lucene parameters
      hibernate.search.backend.type: lucene
      hibernate.search.backend.analysis.configurer: ca.uhn.fhir.jpa.search.HapiHSearchAnalysisConfigurers$HapiLuceneAnalysisConfigurer
      hibernate.search.backend.directory.type: local-filesystem
      hibernate.search.backend.directory.root: target/lucenefiles
      hibernate.search.backend.lucene_version: lucene_current

hapi.fhir.store_resource_in_lucene_index_enabled: true
```

If I send a request like:
```
http://localhost:8080/fhir/Patient?_content:contains=xyz
```
I get back a `400 Bad Request` with the message:
```json
{
    "resourceType": "OperationOutcome",
    "issue": [
        {
            "severity": "error",
            "code": "processing",
            "diagnostics": "HAPI-2566: Fulltext searching is not enabled on this server, can not support the parameter(s): _content"
        }
    ]
}
```

Am I missing something?

henryosei:
You can look at how the application.yml is structured. Try using this snippet below:

hapi.fhir.jpa:
  search:
    allow_multiple_search_results: true
  fulltext_search_enabled: true

Also full-text search support is available in HAPI FHIR v5.0 and above. If you're using an older version, you may need to upgrade to a newer version that supports this feature.

Yes25:
Thank you for the reply @henryosei. 

[...]

PR #828, opened 2025-05-26
feat: add support to enable search index full text

iamsauravsharma:
Add support to enable search index full text disabled by default https://github.com/hapifhir/hapi-fhir/pull/6426. This was enabled by default and non configurable for version 7.6.0 and older and disabled by default from version 8.0.0 and newer which can be enabled
Fixes: #807


robogary:
Formatting check succeeded!

PR #829, opened 2025-06-02
Improve dialect handling

jamesagnew:
The current dialect detection ignores explcitly specified HAPI FHIR dialects

This PR also adds new config flags for a couple of StorageSettings options that weren't exposed.


robogary:
Formatting check succeeded!

PR #845, opened 2025-08-05
Default back to not running with partitions

jkiddo:
@jamesagnew I dont think it was your intention that partitions should be enabled by default, correct?


robogary:
Formatting check succeeded!

PR #861, opened 2025-09-22
Switched to the bitnamilegacy image repository

chgl:
As per: https://github.com/bitnami/containers/issues/83267 soon the bitnami images won't be available. As a temporary workaround I switched to the bitnamilegacy repo which no longer receives updates.

Long-term we should consider switching the tomcat image to the official one and use a different postgres sub-chart. Personally, I migrated several charts to use <https://github.com/CloudPirates-io/helm-charts/tree/main/charts/postgres> which could be an option for here as well.


jkiddo:
Thx!

jkiddo (approved):

PR #960, opened 2026-06-08
Adding support for JS prototyping using GraalJS

jkiddo:
This pull request introduces a new feature that enables secure, server-side execution of administrator-vetted JavaScript scripts for transforming FHIR resources via the new R4 `$execute-javascript` operation. This is implemented using the GraalJS engine, is disabled by default, and includes comprehensive security measures such as sandboxing, strict script selection, and execution timeouts. The PR also adds configuration options and updates documentation and dependencies accordingly.

**New Feature: Server-side JavaScript Execution for FHIR Resources**

- **Implementation of `$execute-javascript` Operation:**
  - Adds `JavaScriptExecutionR4OperationProvider`, a new R4 system-level operation provider that executes administrator-installed JavaScript scripts (not user-supplied code) to transform FHIR resources.
  - Scripts are securely sandboxed (no host, filesystem, or network access) and run with a configurable timeout. Inputs can be inline resources or server-resolved references, and outputs are returned as FHIR resources.

- **Security and Configuration:**
  - Introduces new configuration options in `application.yaml` to enable the feature, specify the scripts directory, and set execution timeouts. The feature is disabled by default and requires explicit configuration to activate.
  - Ensures script names are strictly validated and only files in the configured directory can be executed, preventing path traversal.

**Integration and Dependency Updates**

- **Integration with Server Startup:**
  - Registers the new operation provider in the server configuration, ensuring it is only active when enabled. [[1]](diffhunk://#diff-afcacdb6ba648cc2a2347bd471753b3e4491a011b442f1080961a794542559adR55) [[2]](diffhunk://#diff-afcacdb6ba648cc2a2347bd471753b3e4491a011b442f1080961a794542559adR362) [[3]](diffhunk://#diff-afcacdb6ba648cc2a2347bd471753b3e4491a011b442f1080961a794542559adR428-R429)

- **Dependency Addition:**
  - Adds the GraalJS JavaScript engine dependency (`org.graalvm.js:js`) to the `pom.xml` to enable script execution on stock JDKs.

**Documentation**

- **README Update:**
  - Documents the `$execute-javascript` operation, its security model, configuration, usage examples, and a quickstart for Docker-based testing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -664,6 +664,75 @@ The `$install` operation is triggered with a POST to `[server]/ImplementationGui
 }
 ```
 
+## Server-side JavaScript execution ($execute-javascript)
+
+The R4 system-level operation `$execute-javascript` runs a **server-side** JavaScript file (via the standalone Nashorn engine) to transform FHIR resources. Callers do **not** send code — they reference a script by name that an administrator has placed in a configured directory. The script receives the input resources as a JavaScript array called `input` and returns a single resource object or an array of resource objects, which are returned as `return` parameters.
+
+Security model:
+
+- Disabled by default; enable with `hapi.fhir.javascript_execution_enabled=true`.
+- The `script` parameter is resolved to a file inside `hapi.fhir.javascript_execution_scripts_dir` only (bare file name, no path traversal).
+- Each script runs in a Nashorn sandbox that blocks access to all Java classes (no JVM/filesystem/network reach).
+- Each invocation is bounded by `hapi.fhir.javascript_execution_timeout_seconds` (default `30`); a script that overruns is stopped and the call fails.
+
+Configuration:
+
+```yaml
+hapi:
+  fhir:
+    javascript_execution_enabled: true
+    javascript_execution_scripts_dir: /scripts
+    javascript_execution_timeout_seconds: 30
+```
+
+Inputs may be supplied two ways (combined into `input` in order — inline resources first, then resolved references):
+
+- `resource` (0..*) — an inline FHIR resource.
+- `reference` (0..*) — a literal reference (e.g. `Patient/123`) that the server reads before the script runs.
+
+Example request body to `POST [base]/$execute-javascript`:
+
+```json
+{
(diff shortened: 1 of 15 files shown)


jkiddo:
One _could_ consider to wrap the JS inside FHIR libraries

patrick-werner (approved):

jamesagnew (changes requested):
A couple of initial thoughts - This is a really neat idea, but I don't think this implementation is the right approach:

* We shouldn't be defining functionality in this project - it should purely handle configuration and deployment concerns. If this is a useful feature, it should be built, documented, and tested in a hapi-fhir module, and instantiated in this project.
* It probably makes more sense to use GraalJS for the execution engine - Nashorn isn't being actively developed these days and is fairly behind.

jkiddo:
I hear you on GraalJS (https://github.com/trifork/hapi-fhir-jpaserver-starter/commit/b8f73834e5b7050363356a47aabd827f1718be17) - but then we will also have to require a bump on the Java version to 21 (thats probably fine) soon ™️ .

On moving it to a module to HAPI FHIR ... - I guess... 🤔 - but I wouldn't know the proper encapsulation/abstraction. The reason why I put it here in its current form was because the simple abstraction of just wrapping it in a FHIR operation seemed fine for the purpose. If we move it and just retain (or close to) the signature of 

```java
executeJavascript(
			@OperationParam(name = "script", min = 1, max = 1) StringType theScriptName,
			@OperationParam(name = "resource", min = 0, max = OperationParam.MAX_UNLIMITED)
					List<IBaseResource> theInputResources,
			@OperationParam(name = "reference", min = 0, max = OperationParam.MAX_UNLIMITED)
					List<Reference> theReferences,
			RequestDetails theRequestDetails) 
```

Then I guess that could work just fine.

jkiddo:
Moved to here https://github.com/hapifhir/hapi-fhir/pull/8237 as requested @jamesagnew ✔

PR #975, opened 2026-07-10
feat(helm-chart): add the option to use an HorizontalPodAutoscaler

gfaugere-chut:
Using a simple replicas count is good enough for many use cases, but [HorizontalPodAutoscaler](https://kubernetes.io/docs/concepts/workloads/autoscaling/horizontal-pod-autoscale/) is more desirable when growing to more complex environments

This PR adds it as an option, without changing defaults ofc :)

_Note: I can bundle this with #977 if we want to do a single helm chart release_

### DIFF
--- a/charts/hapi-fhir-jpaserver/templates/deployment.yaml
+++ b/charts/hapi-fhir-jpaserver/templates/deployment.yaml
@@ -9,7 +9,9 @@ metadata:
     {{- toYaml . | nindent 8 }}
   {{- end }}
 spec:
+  {{- if not .Values.autoscaling.enabled }}
   replicas: {{ .Values.replicaCount }}
+  {{- end }}
   selector:
     matchLabels:
       {{- include "hapi-fhir-jpaserver.selectorLabels" . | nindent 6 }}
--- a/charts/hapi-fhir-jpaserver/Chart.yaml
+++ b/charts/hapi-fhir-jpaserver/Chart.yaml
@@ -11,7 +11,7 @@ dependencies:
     repository: oci://registry-1.docker.io/cloudpirates
     condition: postgres.enabled
 appVersion: 8.8.0
-version: 0.24.0
+version: 0.25.0
 annotations:
   artifacthub.io/license: Apache-2.0
   artifacthub.io/containsSecurityUpdates: "false"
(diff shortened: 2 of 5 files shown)


jkiddo:
@chgl will you have a look?

gfaugere-chut:
Also updated to include helm-docs, @jkiddo could you trigger a new CI run please?

gfaugere-chut:
Footer not present in the README as I installed helm-docs from source... I installed the packaged version and reran, should be green now :)

jkiddo:
@chgl it runs green now - I'll merge when you give the 👍

chgl:
Nit picky, but it might be more conventional to place the hpa config behind an `autoscaling` dict, e.g. https://github.com/bitnami/charts/blob/main/bitnami/nginx/templates/hpa.yaml#L6

jkiddo:
@gfaugere-chut if you take take of @chgl 's comment, then I'll merge it

gfaugere-chut:
Hey, sorry for the delay, I've updated the PR to use a simpler to configure `autoscaling` property

charts/hapi-fhir-jpaserver/templates/deployment.yaml
```diff
@@ -9,7 +9,9 @@ metadata:
     {{- toYaml . | nindent 8 }}
   {{- end }}
 spec:
+  {{- if .Values.replicaCount }}
```
chgl: Use 

{{- if not .Values.autoscaling.enabled }}
  replicas: {{ .Values.replicaCount }}
  {{- end }}

Otherwise it's great :)

chgl (changes requested):

charts/hapi-fhir-jpaserver/templates/deployment.yaml
gfaugere-chut: Updated! Cheers

jkiddo (approved):

jkiddo:
@gfaugere-chut - will you take care of the linting bug - then I can merge it

gfaugere-chut:
Looking like a junior here 🙈    
Fixed, hopefuly the test run is smooth (_I would run it locally but it's sloooow on my machine sorry_)

chgl (approved):

PR #977, opened 2026-07-10
feat(helm-chart): add the option to add extra init containers

gfaugere-chut:
The helm chart already gives the ability to add extraVolumes and extraVolumeMounts, which is handy

Sometimes, it is also handy to define other init containers: in my case, I would like to dynamically build a java key store when the container starts, avoiding having to bundle it with a container image

_Note: I can bundle this with #975 if we want to do a single helm chart release_

### DIFF
--- a/charts/hapi-fhir-jpaserver/Chart.yaml
+++ b/charts/hapi-fhir-jpaserver/Chart.yaml
@@ -11,7 +11,7 @@ dependencies:
     repository: oci://registry-1.docker.io/cloudpirates
     condition: postgres.enabled
 appVersion: 8.8.0
-version: 0.23.0
+version: 0.24.0
 annotations:
   artifacthub.io/license: Apache-2.0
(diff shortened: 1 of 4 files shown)


gfaugere-chut:
@chgl you seem to be the "owner" of the helm chart (the main maintainer at least)

Would you be open to review this?

chgl (approved):

chgl:
Sure, I'm fine with reviewing! Unfortunately I can't trigger the workflows to run, but if they pass, LGTM.

chgl:
Ah you'll definitely have to run helm-docs, see the readme.md in the chart to update the docs and commit the changes.

gfaugere-chut:
Updated to include helm-docs, @jkiddo could you trigger a new CI run please?

jkiddo:
@chgl it runs green now - I'll merge when you give the 👍

chgl:
LGTM :)

jkiddo (approved):

PR #984, opened 2026-07-24
Fix MCP search ignoring String query params for Issue#878

vovus893:
RequestBuilder only handled query/searchParams when they came in as a Map, but the MCP tool actually passes them as a String at runtime, so the instanceof check just silently failed and no filter ever got applied. Changed it too just use the query if its a string. Changed the description in ToolFactory too so the query the AI decides on is already in a standard valid format.

### DIFF
--- a/src/main/java/ca/uhn/fhir/jpa/starter/mcp/RequestBuilder.java
+++ b/src/main/java/ca/uhn/fhir/jpa/starter/mcp/RequestBuilder.java
@@ -39,16 +39,10 @@ public class RequestBuilder {
 		switch (interaction) {
 			case SEARCH -> {
 				method = "GET";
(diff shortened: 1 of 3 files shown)


jkiddo:
@vovus893 please add tests for it as well

vovus893:
modified current mcp search test to account for the change

jkiddo:
@vovus893 - I dont see that the changed tests do a query that exercises the modified logic

vovus893:
@jkiddo  added an unfiltered search first so you can
see it's 2 results -> 1 when the query kicks in. also threw in a
RequestBuilderTest that hits buildRequest() directly, skips the
mcp/http stack entirely. lmk if that covers it

